Issue #11, opened 2020-01-15
Split up __init__'s to modules


ryanjdillon:
The __init__ files for the scheduler and executor modules contain class definitions that I think might be navigable as their own modules. The mock definitions could possibly also just live in the tests and be used there. 

While doing this, it could be beneficial to use Abstract Base Classes (i.e. `collections.abs`) to inherit from instead. Ran this by Miles, and he said that a key benefit is that subclasses are automatically enforced to have the methods/interface of the base class.

mrdeveloperdude:
Yes!

This was just beginner-ism on my part, I didn't know what was the recommended Pythonic way.

PR #16, opened 2020-01-17
Update gordo version

ryanjdillon:
Update gordo to the latest `gordo_components` -> `gordo`


ryanjdillon:
This is dependent on #22, so waiting on that one.

app/latigo/gordo/__init__.py
```diff
@@ -346,24 +253,18 @@ def _prepare_auth(self):
...
     def __str__(self):
         return f"GordoModelInfoProvider()"
 
-    def get_models_data(
-        self,
-        projects: typing.Optional[typing.List] = None,
-        model_names: typing.Optional[typing.List] = None,
-    ) -> typing.List[typing.Dict]:
-        models = []
+    def get_machine_data(self, projects: typing.Optional[typing.List] = None, machine_names: typing.Optional[typing.List] = None) -> typing.List[Machine]:
```
mrdeveloperdude: The interface method is called get_model_data() I think. "Machine" is a Gordo specific term, in latigo (and other systems) it is called "Model". I completely agree that it could be named "machine" internally, but I think the interface should remain the same. Thoughts?

ryanjdillon: This most def should stay as you say :+1:  I think this was the result of swathing text replace.

ryanjdillon: Just to reference this in issue #11, implementing the abstract base classes would throw an error once I borked this if the `get_model_data` didn't get implemented.

PR #18, opened 2020-01-20
Implemented auth verification

mrdeveloperdude:
The purpose of this is to fail early when authentication configuration is not correct.

The way it works is by trying to use said auth configuration to contact a presumably valid "test URL" and then fail spectacularly at any hint of connection errors.

### DIFF
--- a/app/latigo/executor/__init__.py
+++ b/app/latigo/executor/__init__.py
@@ -21,10 +21,11 @@ class PredictionExecutor:
     def _fail(self, message: str):
         self.good_to_go = False
         logger.error(message)
-        logger.warning(f"NOTE: Using config:")
-        logger.warning(f"")
-        for line in str(pprint.pformat(self.executor_config)).split("\n"):
-            logger.warning(line)
+        if False:
+            logger.warning(f"NOTE: Using config:")
+            logger.warning(f"")
+            for line in str(pprint.pformat(self.config)).split("\n"):
+                logger.warning(line)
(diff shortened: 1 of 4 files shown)


app/latigo/executor/__init__.py
```diff
@@ -202,10 +209,11 @@ def idle_count(self, has_task):
     def run(self):
         if not self.good_to_go:
             sleep_time = 20
-            logger.error(" ### ### Latigo could not be started!")
-            logger.error("         Please see previous error messages for clues as to why.")
             logger.error("")
+            logger.error(" ### ### Latigo could not be started!")
```
epa095: Note that these kind of log-messages can be problematic later on in an operation setting when one uses log-scraping from the stdout of the docker container. These are all the same event, but they will be printed as seperate log-messages, and there is really no good way of automatically collapsing them to the same event. There are two alternatives:
- Write everything as a single log-message with everything on a single line.
- Write everything as a single log-message with linebreaks in the message (using `""" blabla """` this can be not-that-ugly in python). The latter method is similar to how a stack-trace would be, and it is often possible to configure the log-shipper to figure out that the multiple lines are part of the same event. Also, in python it is possible to configure several formaters simultaneously, so it will print to stdout and also save logs as json in a file. And if you write a single log-message with linebreaks this will be stored as a single log-message in the json-format.

mrdeveloperdude: Enig.

PR #22, opened 2020-01-22
Formatting changes and updates to tests

ryanjdillon:
This is a mixed bag PR of things I was fixing to get tests to run with the update to Gordo, which I've pulled out to its own PR. 
* Use black standard 88 chars for test formating
* Add unit tests for auth module
* Make fixture out of dummy config
* Move MockSensorDataProvider to a fixture in conftest


ryanjdillon:
This is dependent on #22, so waiting for that one.

PR #35, opened 2020-02-14
Add `adal` and `requests` versions to the log output summary.

o-fedorov:
Example of outputs:

```
latigo-executor-1    | Executor settings:
latigo-executor-1    |
latigo-executor-1    |   Good to go:       Yes
latigo-executor-1    |   Latigo Version:   0.1.9
latigo-executor-1    |   Gordo Version:    0.50.0
latigo-executor-1    |   ADAL Version:     1.2.1
latigo-executor-1    |   Requests Version: 2.21.0
latigo-executor-1    |   Restart interval: Disabled
latigo-executor-1    |
```

```
latigo-scheduler     | Scheduler settings:
latigo-scheduler     |
latigo-scheduler     |   Good to go:       Yes
latigo-scheduler     |   Latigo Version:   0.1.9
latigo-scheduler     |   Gordo Version:    0.50.0
latigo-scheduler     |   ADAL Version:     1.2.1
latigo-scheduler     |   Requests Version: 2.21.0
latigo-scheduler     |   Restart interval: Disabled
latigo-scheduler     |   Run at once :     True
latigo-scheduler     |   Start time :      08:00:00
latigo-scheduler     |   Interval:         1 hr, 30 min
latigo-scheduler     |   Data delay:       24 hr
latigo-scheduler     |   Backfill max:     7 day
latigo-scheduler     |   Next start:       2020-02-14 14:00:00 (in 1 hr, 19 min, 47 sec, 769 ms)
latigo-scheduler     |   Projects:         ioc-1901
latigo-scheduler     |
```


app/latigo/log.py
```diff
@@ -1,9 +1,27 @@
...
 
+import adal
+import gordo
+import requests
+
+import latigo
 
 once = False
 
+_SUMMARY_TEMPLATE = f"""
```
mrdeveloperdude: I understand that you want to adhere to the DRY principle here, but I am not sure I would move this log output into log.py. I feel that the log output is too specific to the place from which it is outputted to warrant being re-used. Especially since the formatting of the output now needs to be maintained over two separate locations.

What do you think?

tests/unit/test_log.py
```diff
@@ -0,0 +1,59 @@
+import re
```
mrdeveloperdude: Excellent!

app/latigo/log.py
o-fedorov: Agree. Still It was the first thing came to my mind when I thought about a proper place to add code that generates common log output. I would avoid "generic" places like a `utils` module.

Could you please suggest a better place to add this code?

o-fedorov: Ping @mrdeveloperdude

mrdeveloperdude: I guess this pr is outdated now, sinec it was an exercise to get into the project. But My suggestion would be to perform it the same way as the existing version numbers are fetched, and then later follow up with a refactoring to "generalize" the code for fetching version numbers. I will close this PR now.

PR #38, opened 2020-03-04
use requests-ms-auth==0.0.10 for proper session auth

devProdigy:
# TODO change on higher version after token renew will be fixed


mrdeveloperdude:
It doesn't matter I guess for what you needed.

PR #42, opened 2020-03-13
Fix/gordo bugs

mrdeveloperdude:
Fixed these bugs in Gordo integration. Next up is some more involved stuff that appears to be related to the changes in Gordo Client.

### DIFF
--- a/app/latigo/executor/__init__.py
+++ b/app/latigo/executor/__init__.py
@@ -228,7 +228,7 @@ class PredictionExecutor:
                 )
         except Exception as e:
             logger.error(
-                f"Could not fetch sensor data for task '{task.project_name}.{task.model_name}': {e}",
+                "Could not fetch sensor data for task '%s.%s':", task.project_name, task.model_name,
                 exc_info=True,
             )
         return sensor_data
--- a/app/latigo/executor/__init__.py
+++ b/app/latigo/executor/__init__.py
@@ -219,18 +219,18 @@ class PredictionExecutor:
                     spec, time_range
                 )
                 if not sensor_data:
-                    logger.warning(f"Error getting sensor data: {err}")
+                    logger.error(f"Error getting sensor data: {err}")
                 elif not sensor_data.ok():
-                    logger.warning(f"Sensor data '{sensor_data}' was not ok")
+                    logger.error(f"Sensor data '{sensor_data}' was not ok")
             else:
-                logger.warning(
+                logger.error(
                     f"Error getting spec for project={project_name} and model={model_name}"
                 )
         except Exception as e:
             logger.error(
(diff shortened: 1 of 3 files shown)


app/latigo/gordo/misc.py
```diff
@@ -67,20 +67,21 @@ def print_client_debug(client: typing.Optional[Client]):
...
+            "data_provider": client.data_provider,
+            "use_parquet": client.use_parquet,
+            "session": client.session,
+            "prediction_path": client.prediction_path,
+            "batch_size": client.batch_size,
+            "parallelism": client.parallelism,
+            "forward_resampled_sensors": client.forward_resampled_sensors,
+            "n_retries": client.n_retries,
+        }
+    except:
```
o-fedorov:
```suggestion
    except Exception:
```
Bare `except` is considered bad practice.

app/latigo/gordo/misc.py
```diff
@@ -67,20 +67,21 @@ def print_client_debug(client: typing.Optional[Client]):
...
+        logger.error(f"Could not log client {pprint.pformat(client)}", exc_info=True)
     logger.info(pprint.pformat(data))
```
o-fedorov:
```suggestion
    else:
        logger.info(pprint.pformat(data))
```

Log data content only if it was successfully built

app/latigo/executor/__init__.py
```diff
@@ -219,18 +219,18 @@ def _fetch_sensor_data(self, task: Task) -> typing.Optional[SensorDataSet]:
...
+                    logger.error(f"Sensor data '{sensor_data}' was not ok")
             else:
-                logger.warning(
+                logger.error(
                     f"Error getting spec for project={project_name} and model={model_name}"
                 )
         except Exception as e:
             logger.error(
-                f"Could not fetch sensor data for task '{task.project_name}.{task.model_name}': {e}"
+                f"Could not fetch sensor data for task '{task.project_name}.{task.model_name}': {e}",
```
o-fedorov:
```suggestion
                "Could not fetch sensor data for task '%s.%s':", task.project_name, task.model_name,
```
Logging documentation suggests delayed strings substitution, and pass the values as parameters. Also there is no need to add the exception instance to the message since with `exc_info=True` it will be provided it anyway.

app/latigo/gordo/model_info_provider.py
```diff
@@ -27,6 +27,7 @@
 
 from .misc import *
 from .client_pool import *
+from .data_provider import _gordo_to_latigo_tag_list
```
o-fedorov: Private functions (ones that start with an underscore) should not be used outside their module. That may lead to bugs in case of refactoring.

mrdeveloperdude: I am aware. Will fix this in a later PR

PR #48, opened 2020-03-23
Feature/us 09 configure auth to metadata api

devProdigy:
* Add Client for connecting to the Metadata API
* Enabled integration Metadata API tests on the CI step
* bumped version of the Session auth library to use auto headers "Ocp-Apim-Subscription-Key"


app/latigo/metadata_api/client.py
```diff
@@ -0,0 +1,66 @@
...
+            return self._raise_exception("No base_url found in config")
+
+    def _create_session(self, force: bool = False):
+        """Create session for the future calls."""
+        if not self.session or force:
+            self.session = requests_ms_auth.MsRequestsSession(
+                requests_ms_auth.MsSessionConfig(**self.auth_config)
+            )
+        if not self.session:
+            return self._raise_exception(f"Could not create session for {self.__class__.__name__}")
```
o-fedorov: Nit-pick:
```suggestion
            return self._raise_exception(f"Could not create session for { type(self).__name__} ")
```

app/latigo/metadata_api/client.py
```diff
@@ -0,0 +1,66 @@
...
+    @staticmethod
+    def _raise_exception(message: str, exception_type=Exception) -> None:
+        """Raise passed exception type with the message. Write message to the error log."""
+        logger.error(message)
+        raise exception_type(message)
+
+    def get(self, *args, **kwargs) -> Response:
+        """Make GET call the the API. Standard params are allowed."""
+        res = self.session.get(*args, **kwargs)
+        return res
```
o-fedorov:
```suggestion
        return self.session.get(*args, **kwargs)
```

app/latigo/metadata_api/client.py
```diff
@@ -0,0 +1,66 @@
...
+        if not self.session:
+            return self._raise_exception(f"Could not create session for {self.__class__.__name__}")
+
+    @staticmethod
+    def _raise_exception(message: str, exception_type=Exception) -> None:
+        """Raise passed exception type with the message. Write message to the error log."""
+        logger.error(message)
+        raise exception_type(message)
+
+    def get(self, *args, **kwargs) -> Response:
```
o-fedorov: Could you please provide the arguments that may be required explicitly? Here and below.

o-fedorov:
I see that having a class `MetadataAPIClient` adds unnecessary complexity, since it acts as a tiny wrapper of `requests_ms_auth.MsRequestsSession`.

It would be simpler to just have a function that builds and returns `requests_ms_auth.MsRequestsSession` directly.

Also this code may belong to `requests_ms_auth` itself, so that it's better to move it there.

o-fedorov:
Also do the tests perform *real* requests to an external service? If it is true, it may be bad practice for multiple reasons.

o-fedorov:
@devProdigy , @mrdeveloperdude  ⬆️

PR #55, opened 2020-04-01
Feature/us 09 send data latigo to metadata api

devProdigy:
* metadata storage provider interface;
* adjust 'name' in data structure for Metadata API;
* implement metadata storage provider with code refactoring;
* add common project quality settings ;
* adjust existing tests with name changes.

### DIFF
--- a/app/latigo/executor/__init__.py
+++ b/app/latigo/executor/__init__.py
@@ -255,8 +255,7 @@ class PredictionExecutor:
 
         except Exception as e:
             logger.error(
-                f"Could not store prediction or metadata data for task '{task.project_name}.{task.model_name}, "
-                f"time period: from '{task.from_time}' to '{task.to_time}''. Error: {e}"
+                f"Could not store prediction or metadata for task {self.make_prediction_task_info(task)}. Error: {e}."
             )
             raise e
 
@@ -337,6 +336,7 @@ class PredictionExecutor:
                             f"{task.from_time} lasting {task.to_time - task.from_time} for '{task.model_name}' "
                             f"in '{task.project_name}'"
                         )
+                        logger.info(f"[Prediction_task_info] {self.make_prediction_task_info(task)}")
 
                         revision = self.model_info_provider.get_project_latest_revisions(task.project_name)
 
@@ -357,8 +357,7 @@ class PredictionExecutor:
                             except InsufficientDataAfterRowFilteringError as e:
                                 logger.warning(
                                     "[Skipping the prediction 'InsufficientDataAfterRowFilteringError']: "
-                                    f"'{task.project_name}.{task.model_name}, "
-                                    f"time period: from '{task.from_time}' to '{task.to_time}''. Error: {e}"
+                                    f"{self.make_prediction_task_info(task)}"
                                 )
 
                             if prediction_data and prediction_data.ok():
@@ -367,7 +366,7 @@ class PredictionExecutor:
                                     datetime.datetime.now() - task_fetch_start
                                 )
                                 logger.info(
-                                    f"Prediction stored after {human_delta(prediction_storage_interval)}"
+                                    f"Prediction stored after {human_delta(prediction_storage_interval)}\n"
                                 )
                                 self.idle_count(True)
                             else:
@@ -398,3 +397,8 @@ class PredictionExecutor:
             # logger.info("Executor stopped processing")
         else:
             logger.error("No task queue")
+
+    @staticmethod
+    def make_prediction_task_info(task: Task) -> str:
+        """Make info about prediction task for logging."""
+        return f"'{task.project_name}.{task.model_name}', prediction: from '{task.from_time}' to '{task.to_time}'"
--- a/app/latigo/executor/__init__.py
+++ b/app/latigo/executor/__init__.py
@@ -284,14 +284,14 @@ class PredictionExecutor:
                 meta, error = self.prediction_storage_provider._get_meta_by_name(name=tag.name, asset_id=tag.asset)
                 if error:
                     # Raise error if tag does not exist
-                    raise Exception(f"Tag was not found. Name='{tag.name}' asset='{tag.asset}': error='{error}'")
+                    raise ValueError(f"Tag was not found. Name='{tag.name}' asset='{tag.asset}': error='{error}'")
 
                 tag_id = get_time_series_id_from_response(meta)
                 input_time_series_ids[tag.name] = tag_id
 
         missing_tags = {key: val for key, val in input_time_series_ids.items() if not val}
         if missing_tags:
-            raise Exception("[TAG_NOT_FOUND]: " + ';'.join(missing_tags.keys()))
+            raise ValueError("[TAG_NOT_FOUND]: " + ';'.join(missing_tags.keys()))
         return input_time_series_ids
 
     def idle_count(self, has_task):
@@ -357,7 +357,7 @@ class PredictionExecutor:
                             except InsufficientDataAfterRowFilteringError as e:
                                 logger.warning(
                                     "[Skipping the prediction 'InsufficientDataAfterRowFilteringError']: "
-                                    f"{self.make_prediction_task_info(task)}"
+                                    f"{self.make_prediction_task_info(task)}. Error: {e}"
                                 )
 
                             if prediction_data and prediction_data.ok():
--- a/app/latigo/executor/__init__.py
+++ b/app/latigo/executor/__init__.py
@@ -8,6 +8,9 @@ from gordo import __version__ as gordo_version
 from requests_ms_auth import __version__ as auth_version
 
 from latigo.gordo import NoTagDataInDataLake
+from latigo.metadata_storage import prediction_metadata_storage_provider_factory
+from latigo.time_series_api import get_time_series_id_from_response
+from latigo.time_series_api.misc import MODEL_INPUT_OPERATION
 from latigo.types import (
     Task,
     SensorDataSpec,
@@ -34,6 +37,7 @@ class PredictionExecutor:
         self._prepare_task_queue()
         self._prepare_sensor_data_provider()
         self._prepare_prediction_storage_provider()
+        self._prepare_prediction_metadata_storage_provider()
         self._prepare_model_info()
         self._prepare_prediction_executor_provider()
         self._prepare_executor()
@@ -86,6 +90,20 @@ class PredictionExecutor:
         )
         if not self.prediction_storage_provider:
             self._fail(f"No prediction storage configured, cannot continue...")
+
+    # Inflate prediction metadata storage provider from config
+    def _prepare_prediction_metadata_storage_provider(self):
+        """Initialize provider for storing of prediction metadata."""
+        prediction_metadata_storage_provider_config = self.config.get(
+            "prediction_metadata_storage", None
+        )
+        if not prediction_metadata_storage_provider_config:
+            self._fail("No prediction_metadata_storage specified")
+        self.prediction_metadata_storage_provider = prediction_metadata_storage_provider_factory(
+            prediction_metadata_storage_provider_config
+        )
+        if not self.prediction_metadata_storage_provider:
+            self._fail(f"No prediction metadata storage configured, cannot continue...")
 
     # Inflate prediction executor provider from config
     def _prepare_prediction_executor_provider(self):
@@ -141,7 +159,7 @@ class PredictionExecutor:
             f"  Restart interval: {restart_interval_desc}\n"
         )
 
-    def _fetch_spec(self, project_name: str, model_name: str):
+    def fetch_spec(self, project_name: str, model_name: str) -> SensorDataSpec:
         return self.model_info_provider.get_spec(
             project_name=project_name, model_name=model_name
         )
@@ -170,7 +188,7 @@ class PredictionExecutor:
             time_range: TimeRange = TimeRange(task.from_time, task.to_time)
             project_name: str = task.project_name
             model_name: str = task.model_name
-            spec: SensorDataSpec = self._fetch_spec(project_name, model_name)
+            spec: SensorDataSpec = self.fetch_spec(project_name, model_name)
             if spec:
                 sensor_data, err = self.sensor_data_provider.get_data_for_range(
                     spec, time_range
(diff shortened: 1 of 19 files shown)


app/latigo/executor/__init__.py
```diff
@@ -87,6 +91,20 @@ def _prepare_prediction_storage_provider(self):
         if not self.prediction_storage_provider:
             self._fail(f"No prediction storage configured, cannot continue...")
 
+    # Inflate prediction metadata storage provider from config
```
o-fedorov, line 94: Should this comment go to the docstring instead?

code_quality/Makefile
```diff
@@ -22,8 +22,8 @@ mypy:
 	mypy --ignore-missing-imports "${TESTS_DIR}"  || true
 
 black:
-	black -l 88 -t py37 "${APP_DIR}"
-	black -l 88 -t py37 "${TESTS_DIR}"
+	black -l 120 -t py37 "${APP_DIR}"
```
o-fedorov, line 25: These values should go to the config file.

app/latigo/metadata_storage/__init__.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+def prediction_metadata_storage_provider_factory(config):
+    provider_type = config.get("type", None)
+
+    if "metadata_api" == provider_type:
+        from latigo.metadata_api.metadata_storage_provider import MetadataAPIMetadataStorageProvider
+
+        provider = MetadataAPIMetadataStorageProvider(config)
+    elif "mock" == provider_type:
```
o-fedorov, line 61: Implementation should not depend on tests. Use the provider explicitly and use `unittest.mock.patch` to replace it with a mock one for tests.

app/latigo/time_series_api/misc.py
```diff
@@ -101,20 +82,23 @@ def _parse_request_json(
     return None, "ERRROR"
 
 
-invalid_operations = ["start", "end", "model-input"]
+MODEL_INPUT_OPERATION = "model-input"
+INVALID_OPERATIONS = ["start", "end", MODEL_INPUT_OPERATION]
```
o-fedorov, line 86:
```suggestion
INVALID_OPERATIONS = frozenset(["start", "end", MODEL_INPUT_OPERATION])
```

app/latigo/metadata_api/metadata_storage_provider.py
```diff
@@ -0,0 +1,85 @@
...
+                    OutputTag(
+                        name=output_tag_names[col],
+                        time_series_id=output_time_series_ids[(operation, tag_name)],
+                        type=make_output_tag_type(tag_name),
+                        derived_from=make_output_tag_derived_from(tag_name),
+                        description=make_output_tag_description(operation, tag_name),
+                    )
+                )
+            else:
+                continue
```
o-fedorov, line 66:
```suggestion
            if operation not in INVALID_OPERATIONS:
                continue

            if operation == MODEL_INPUT_OPERATION:
                tag_time_series_id = input_time_series_ids[tag_name]
                metadata_api_input_tags.append(InputTag(name=tag_name, time_series_id=tag_time_series_id))
            else:
                metadata_api_output_tags.append(
                    OutputTag(
                        name=output_tag_names[col],
                        time_series_id=output_time_series_ids[(operation, tag_name)],
                        type=make_output_tag_type(tag_name),
                        derived_from=make_output_tag_derived_from(tag_name),
                        description=make_output_tag_description(operation, tag_name),
                    )
                )
```
Use already defined constants, follow Short-Circuit Break pattern.

app/latigo/metadata_api/metadata_storage_provider.py
```diff
@@ -0,0 +1,85 @@
...
+
+        project_name = prediction_data.meta_data.project_name
+        revision = prediction_data.meta_data.revision
+        model_name = prediction_data.meta_data.model_name
+        model_training_period = prediction_data.meta_data.model_training_period
+
+        # fill the tags
+        for col in df_columns:
+            operation = col[0]  # example: "start", "end", "model-input"
+            tag_name = col[1]  # example: "1903.R-29TT3018.MA_Y"
```
o-fedorov, line 51:
```suggestion
            operation, tag_name, *_ = col
```

app/latigo/metadata_api/metadata_storage_provider.py
```diff
@@ -0,0 +1,85 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class MetadataAPIMetadataStorageProvider(MetadataAPIClient, MetadataStorageProviderInterface):
+    def __init__(self, config: dict):
+        super().__init__(config)
+
+    def __str__(self):
+        return f"{self.__class__.__name__}({self.base_url})"
```
o-fedorov, line 19: This representation looks misleading, since `MetadataAPIMetadataStorageProvider(base_url)` would raise an error. Use angular braces to indicate that string representation can't be used to initialize the class.
```suggestion
        return f"<{ type(self).__name__ }: { self.base_url }>"
```

app/latigo/metadata_api/metadata_storage_provider.py
```diff
@@ -0,0 +1,85 @@
...
+from latigo.metadata_storage import MetadataStorageProviderInterface
+from latigo.time_series_api.misc import make_output_tag_derived_from, make_output_tag_description, make_output_tag_type, \
+    make_prediction_metadata_description
+from latigo.types import PredictionDataSet
+
+logger = logging.getLogger(__name__)
+
+
+class MetadataAPIMetadataStorageProvider(MetadataAPIClient, MetadataStorageProviderInterface):
+    def __init__(self, config: dict):
```
o-fedorov, line 15: This method is redundant.

app/latigo/metadata_api/metadata_storage_provider.py
```diff
@@ -0,0 +1,85 @@
...
+            input_tags=metadata_api_input_tags,
+            output_tags=metadata_api_output_tags,
+        )
+        res = self.send_time_series_id_metadata(ts_metadata)
+        if res.status_code != 200:
+            raise Exception(f"[METADATA_STORING_ERROR]: {res.status_code} - {res.text}.")
+
+        logger.info(f"[MODEL_ID] Prediction metadata was stored to the Metadata API. "
+                    f"Record ID - '{res.json()['model_id']}'")
+        return None
```
o-fedorov, line 85: This return statement is redundant.

app/latigo/executor/__init__.py
devProdigy: for this release I've integrate code with using the same structure as was in the project previously. After it will work properly this should be refactored definitely, but for the sake of keeping similarity it has such "things".

code_quality/Makefile
devProdigy: I've already placed it to the root of the project to the proper config files but cause of the way as makefile works it won't see them from this dir. This will be moved in the next versions as well.

app/latigo/metadata_storage/__init__.py
devProdigy: Totally agree.
But for this release I've integrate code with using the same structure as was in the project previously. After it will work properly this should be refactored definitely, but for the sake of keeping similarity it has such "things".

app/latigo/time_series_api/misc.py
```diff
@@ -124,3 +108,66 @@ def prediction_data_naming_convention(
     operation = operation.replace(separator, replacement)
     return f"{tag_name}{separator}{model_name}{separator}{operation}"
 
+
```
o-fedorov, line 111: I wonder why most of the functions below ever exist. They are just single-liners that are used in one single place. Reading code like this forces to jump back and forth to just see that a function wraps one simple expression.

app/latigo/metadata_api/metadata_storage_provider.py
devProdigy: thanks for pointing to the constants that I've forgot to add, I'll replace them. But with such change logic will be wrong - we should skip only ["start", "end"] but process all other.

o-fedorov: Sure, it should be

```
            if operation in INVALID_OPERATIONS and operation != MODEL_INPUT_OPERATION:
                continue
```
Or define a separate constant for {start, stop}.

app/latigo/time_series_api/misc.py
devProdigy: at least for testing purposes. That logic has been changed for the last few days few times, so I'm gonna write separate tests for it, so this business logic will be explicit and won't be accidentally changed.

app/latigo/time_series_api/misc.py
devProdigy: nice, done

app/latigo/metadata_api/metadata_storage_provider.py
devProdigy: done

app/latigo/time_series_api/misc.py
o-fedorov: Still, I can't accept this architectural choice. These functions code should be inlined where they are called, and then the resulting code should be split to logical parts.

For example, having a bunch of `make_output_tag_*` functions suggests that their code actually belongs to `OutputTag` class.

devProdigy: Moving `make_output_tag_*` to the `OutputTag` class sounds logically to me. Also, move `make_prediction_metadata_description` to class `TimeSeriesIdMetadata` will make sense as well.
Will it be ok if I move them there?

o-fedorov: Sure

devProdigy: done, please resolve if it's ok

app/latigo/executor/__init__.py
```diff
@@ -214,18 +237,62 @@ def _execute_prediction(
...
+
+            if operation == MODEL_INPUT_OPERATION:
+                # save tag names for further fetching its Time Series IDs
+                input_time_series_ids[tag_name] = ""
+
+        spec = self.fetch_spec(prediction_data.meta_data.project_name, prediction_data.meta_data.model_name)
+
+        for tag in spec.tag_list:
+            if tag.name in input_time_series_ids.keys():
+                meta, error = self.prediction_storage_provider._get_meta_by_name(name=tag.name, asset_id=tag.asset)
```
o-fedorov, line 284: Private methods should not be accessed outside of their class.

app/latigo/executor/__init__.py
```diff
@@ -214,18 +237,62 @@ def _execute_prediction(
...
+                if error:
+                    # Raise error if tag does not exist
+                    raise Exception(f"Tag was not found. Name='{tag.name}' asset='{tag.asset}': error='{error}'")
```
o-fedorov, line 287: You should consider raising more specific errors. `TypeError`, `ValueError`, `LookupError` and `RuntimeError` are good candidates.

Here and in other places.

app/latigo/executor/__init__.py
o-fedorov: It is what is called "broken windows": if windows around are broken, people tend to break more ones.  Just remove the comment as it is redundant and there is already a docstring.

app/latigo/metadata_api/client.py
```diff
@@ -68,3 +67,18 @@ def post(self, *args, **kwargs) -> Response:
...
+
+    def send_time_series_id_metadata(self, time_series_ids_metadata: TimeSeriesIdMetadata) -> Response:
+        """Save time series id metadata to the Metadata API after storing it in the Time Series API."""
+        url = f"{self.base_url}/models"
+
+        res = self.post(url=url, data=self._dump_data(asdict(time_series_ids_metadata)))
+        return res
+
+    @staticmethod
+    def _dump_data(data: Any) -> str:
```
o-fedorov, line 79:
```suggestion
    def _dump_data(data: Dict) -> str:
```
According to the docstring.

app/latigo/metadata_api/data_structures.py
```diff
@@ -22,25 +24,47 @@ class OutputTag:
...
+    @staticmethod
+    def make_output_tag_derived_from(tag_name: str) -> Union[str, None]:
+        """Make 'derived_from' for output_tag metadata."""
+        return tag_name or None
+
+    @staticmethod
+    def make_output_tag_description(operation: str, tag_name: str) -> str:
+        """Make description for output_tag."""
+        return f"Gordo {operation} - {tag_name}"
+
```
o-fedorov, line 67: The methods should not be static.  The dataclass already have methods `type`, `description` and `derived_from`.  Use the code to `__post_init__` to set these attributes if they are not already set.

app/latigo/metadata_api/data_structures.py
```diff
@@ -69,3 +93,12 @@ class TimeSeriesIdMetadata:
     description: str = None
     status: str = "not_defined"
     labels: List[str] = None
+
+    @staticmethod
```
o-fedorov, line 97: The same here.  Update the dataclass' __post_init__ to set the description using other dataclass paramenters.

app/latigo/metadata_storage/__init__.py
```diff
@@ -0,0 +1,65 @@
+import logging
```
o-fedorov, line 1: Do not put business logic to `__init__.py`. For simple modules use single file (`app/latigo/metadata_storage.py`), if the module needs decomposition, use a package with a dedicated file inside it, like so:

```
app/latigo/metadata_storage/
|- __init__.py  # "Glue" code.
|_ <module_name>.py  # Business logic
```

app/latigo/time_series_api/misc.py
```diff
@@ -124,3 +107,38 @@ def prediction_data_naming_convention(
...
+        '1903.R-29PST3037.MA_Y' -> '1903'. Split first non empty tag name with '.' and take first part as asset.
+    """
+    for operation, tag_name in columns:
+        if tag_name:
+            return tag_name.split(".")[0]
+    raise Exception(
+        f"Not common asset found in dataframe columns. Parsed data: " + "; ".join(col[1] for col in columns.values)
+    )
+
+
```
o-fedorov, line 124: The code below belongs to `TimeSeriesAPIClient`.  Also since it's common for the client to pass around `meta, error`, with searching for `time_series_id` in `meta`, consider providing a [typing.NamedTuple](https://docs.python.org/3.7/library/typing.html#typing.NamedTuple) for (`meta: Optional[dict], error: Optional[str]`), and add the search functionality to the namedtuple itself.

Examples where `meta, error` and `get_time_series_id_from_response(meta)` are used sequentially: 

* https://github.com/equinor/latigo/blob/f64fcbc97a19dc3b5e1aea9528035c9e88146e35/app/latigo/time_series_api/prediction_storage_provider.py#L68
* https://github.com/equinor/latigo/pull/55/files#diff-ab28f04a938ce4fa424e4aee005621ddR284

o-fedorov:
I don't want to block the PR any more.  Take the comments into consideration and resolve them on your own.

PR #58, opened 2020-04-03
adjust pre-prediction-data filtering

devProdigy:
* added and enabled few unit tests on CI


app/latigo/utils.py
```diff
@@ -282,7 +282,8 @@ def local_datetime_to_utc(target: datetime) -> str:
          String representation of without and taking into the account time zone formatted in UTC timezone.
             Do not make changes if already in UTC timezone or does not have any.
     """
-    if target.tzinfo is None or dateutil.tz.tzutc() or target.tzinfo == pytz.utc:
+    if target.tzinfo is None or not target.utcoffset() or target.tzinfo == pytz.utc:
+        target = target.replace(tzinfo=None)
```
o-fedorov, line 286: It would be bette to set `tzinfo` to `utc` to not mix tz-naive and tz-aware datetimes.

tests/unit/test_utils.py
```diff
@@ -1,13 +1,19 @@
...
+from datetime import datetime
+from unittest import mock
+
+import pytest
+import pytz
+
 from latigo.utils import (
     merge,
-    load_config,
+    # load_config,
```
o-fedorov, line 12: Just remove it.

PR #59, opened 2020-04-10
US-09 fix timezone adjustments for Kafka messages

devProdigy:
## Description

Previously local datetime was transformed to the timestamp for Kafka messages for `from_time` and `to_time` without taking into account timezone, thus final time was distorted.

## Type of change

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

[Azure DevOps Link](https://dev.azure.com/EquinorASA/OMNIA%20Prevent/_sprints/taskboard/OMNIA%20Prevent%20Team/OMNIA%20Prevent/Sprint%207?workitem=16578)


app/latigo/scheduler/__init__.py
```diff
@@ -158,13 +158,12 @@ def perform_prediction_step(self):
...
-            datetime.datetime.now() - self.continuous_prediction_delay
-        )
-        prediction_end_time = (
-            prediction_start_time + self.continuous_prediction_interval
-        )
+        now_utc = get_datetime_now_in_utc()
+        stats_start_time = now_utc
+        # 'prediction_start_time' and 'prediction_end_time' should be in UTC time cause Queue will ignore timezone.
+        prediction_start_time_utc = now_utc - self.continuous_prediction_delay
+        prediction_end_time_utc = prediction_start_time_utc + self.continuous_prediction_interval
```
o-fedorov, line 165:
```suggestion
        stats_start_time = get_datetime_now_in_utc()
        # 'prediction_start_time' and 'prediction_end_time' should be in UTC time cause Queue will ignore timezone.
        prediction_start_time_utc = stats_start_time - self.continuous_prediction_delay
        prediction_end_time_utc = prediction_start_time_utc + self.continuous_prediction_interval
```
There is no need to have 2 variables at the same place, and that contain the same value.

o-fedorov:
I would like to see at least one test that validates the fix.

devProdigy:
> I would like to see at least one test that validates the fix.

Added but it took more time cause there's no working tests for the whole scheduler.

tests/unit/test_scheduler.py
```diff
@@ -0,0 +1,39 @@
...
+
+@pytest.fixture
+@patch("latigo.task_queue.kafka.Producer", new=MagicMock())
+@patch("latigo.scheduler.Scheduler._prepare_model_info", new=MagicMock())
+@patch("latigo.scheduler.Scheduler._perform_auth_checks", new=MagicMock())
+def scheduler(schedule_config, monkeypatch) -> Scheduler:
+    monkeypatch.setattr("latigo.scheduler.get_datetime_now_in_utc", MagicMock(return_value=DATETIME_UTC_NOW))
+    scheduler = Scheduler(schedule_config)
+    scheduler.models = MODELS
+    return scheduler
```
o-fedorov, line 22:
```suggestion
@pytest.fixture
@patch("latigo.task_queue.kafka.Producer")
@patch("latigo.scheduler.Scheduler._prepare_model_info")
@patch("latigo.scheduler.Scheduler._perform_auth_checks")
@patch("latigo.scheduler.get_datetime_now_in_utc", return_value=DATETIME_UTC_NOW)
def scheduler(schedule_config) -> Scheduler:
    scheduler = Scheduler(schedule_config)
    scheduler.models = MODELS
    return scheduler
```
1. There is a mix of `monkeypatch` and `unittest.mock` libraries usage.  I would suggest to use just one.
2. Unless there are dynamically set public properties, it's more safe to let `patch` use auto-generated spec of a patched object.

tests/unit/test_scheduler.py
```diff
@@ -0,0 +1,39 @@
...
+from latigo.scheduler import Scheduler
+from tests.conftest import SCHEDULER_PREDICTION_DELAY, SCHEDULER_PREDICTION_INTERVAL
+
+DATETIME_UTC_NOW = datetime.fromisoformat("2020-04-10T10:00:00.000000+00:00")
+MODELS = [Model(model_name="model", project_name="project", tag_list=[], target_tag_list=[])]
+
+
+@pytest.fixture
+@patch("latigo.task_queue.kafka.Producer", new=MagicMock())
+@patch("latigo.scheduler.Scheduler._prepare_model_info", new=MagicMock())
```
o-fedorov, line 16: A note: having to patch private attributes is a clear sign that the code is not testable.  I assume that either the code of these methods should be extracted in another place, or the instances these methods operates on should be mocked instead.

tests/unit/test_scheduler.py
```diff
@@ -0,0 +1,39 @@
...
+    scheduler.models = MODELS
+    return scheduler
+
+
+def test_perform_prediction_step_put_task(scheduler: Scheduler):
+    """Validates task serialisation and using UTC time."""
+    send_event_mock = MagicMock()
+
+    scheduler.task_queue.send_event = send_event_mock
+    scheduler.perform_prediction_step()
```
o-fedorov, line 30: I suppose that `scheduler.task_queue` is already mocked in `scheduler` fixture above, so this code is redundant.

o-fedorov: If it is _not mocked_, then let `mock.patch` generate the spec of a patched object, as noted above.

```suggestion
    with patch.object(scheduler.task_queue, "send_event") as send_event_mock:
        scheduler.perform_prediction_step()
```

tests/unit/test_scheduler.py
devProdigy: monkeypatch is used cause `get_datetime_now_in_utc` is imported before we can apply `patch`

tests/unit/test_scheduler.py
devProdigy: A lot of code should be refactored. For this should be made separate task.

tests/unit/test_scheduler.py
devProdigy: nice, forgot about this
thanks

PR #62, opened 2020-04-14
US-09 update spec and sent metadata

devProdigy:
This is update for the further update of the Metadata API spec.
Changes are backward compatible.


app/latigo/metadata_api/data_structures.py
```diff
@@ -39,8 +39,8 @@ class OutputTag:
     name: str
     time_series_id: str
     type: str
-    description: str = None
-    derived_from: str = None
+    description: Optional[str] = ""
+    derived_from: Optional[str] = ""
```
o-fedorov, line 43: `Optional` means that it may be `None`. Is it intended? I suppose it is always `str`:
```suggestion
    description: str = ""
    derived_from: str = ""
```

devProdigy: nice catch, I'll roll logic in other way

devProdigy: done

PR #63, opened 2020-04-21
feat: add trivy scan step

r-t-m:
## Description

Summary of the changes:
- Adds docker image scanning step with trivy.
- Branch check condition removed in the build step to get feedback on found vulnerabilities and to act on that before push to master.

## Type of change

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes
- [x] CI related changes
- [ ] Tests
- [ ] Other

# Checklist:

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -15,12 +15,18 @@ env:
   DOCKER_USERNAME: ${{secrets.DOCKER_USERNAME}}
   DOCKER_PASSWORD: ${{secrets.DOCKER_PASSWORD}}
   DOCKER_REPO: ${{secrets.DOCKER_REPO}}
-
 jobs:
   build:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v1
+    - name: free disk space
+      run: |
+        sudo swapoff -a
+        sudo rm -f /swapfile
+        sudo apt clean
+        docker rmi $(docker image ls -aq)
+        df -h
     - name: List environment info
       run: |
         make show-env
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -15,12 +15,18 @@ env:
   DOCKER_USERNAME: ${{secrets.DOCKER_USERNAME}}
   DOCKER_PASSWORD: ${{secrets.DOCKER_PASSWORD}}
   DOCKER_REPO: ${{secrets.DOCKER_REPO}}
-
 jobs:
   build:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v1
+    - name: free disk space
```
devProdigy, line 23: can you please explain why do we need to clear space in GitHub Action step?

.github/workflows/main.yml
```diff
@@ -62,9 +68,13 @@ jobs:
         METADATA_API_APIM_KEY: ${{ secrets.METADATA_API_APIM_KEY }}
 
     - name: Build docker images for scheduler and executor
-      if: github.ref == 'refs/heads/master'
```
devProdigy, line 65: this should not be removed. We want this steps to be run only for merging to master.

.github/workflows/main.yml
r-t-m: Scan failed for me one time with the error that there are no disk space left on device.

.github/workflows/main.yml
r-t-m: I removed contition so that you could act on scan output earlier. Are you sure you want it to run only for master?

.github/workflows/main.yml
```diff
@@ -62,9 +68,13 @@ jobs:
...
       run: |
         make build
+
+    - name: Scan image
```
devProdigy, line 74: please add here run only for master branch:
`if: github.ref == 'refs/heads/master'`
This is needed cause it takes more than 4 mins and it's not needed to increase time of CI step for all branches builds (only needed for master branch).

.github/workflows/main.yml
devProdigy: yes, and `name: Scan image` step should be run only on master as well.
No need to build image for other branches.

.github/workflows/main.yml
devProdigy: ok

PR #72, opened 2020-05-13
remove redundant sensor fetch and bump 0.2.10 -> 0.2.11

devProdigy:
## Description

Time execution optimization by removing redundant sensor data call.

[Azure DevOps Link](https://dev.azure.com/EquinorASA/OMNIA%20Prevent/_workitems/edit/17069)

## Type of change

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] CI related changes
- [X] Tests
- [ ] Other

# Checklist:


tests/unit/test_executor.py
```diff
@@ -1,12 +1,24 @@
...
+@patch("latigo.executor.PredictionExecutor._perform_auth_checks", new=MagicMock())
+def test__execute_prediction_success(config):
+    executor = PredictionExecutor(config=config)
+    task = TaskFactory()
+    revision = "000"
 
-def test_fetch_task(config):
-    pe = PredictionExecutor(config=config)
+    client_mock = MagicMock(name="client_mock")
+    with patch.object(executor.prediction_executor_provider.gordo_pool, "allocate_instance", return_value=client_mock):
```
o-fedorov, line 19:
```suggestion
    with patch.object(executor.prediction_executor_provider.gordo_pool, "allocate_instance") as mock_pool:
        client_mock = mock_pool.return_value
```

tests/unit/test_executor.py
```diff
@@ -1,12 +1,24 @@
...
+        executor._execute_prediction(task=task, revision=revision)
```
o-fedorov, line 20: Tasting private methods looks wrong. It seems that `GordoPredictionExecutionProvider.execute_prediction` should be tested instead.

tests/unit/test_executor.py
```diff
@@ -1,12 +1,24 @@
...
+from latigo.gordo import GordoModelInfoProvider
+from tests.factories.task import TaskFactory
 
-logger = logging.getLogger(__name__)
 
+@patch("latigo.gordo.prediction_execution_provider.GordoClientPool", new=MagicMock())
+@patch(
+    "latigo.executor.model_info_provider_factory", new=MagicMock(side_effect=MagicMock(spec_set=GordoModelInfoProvider))
+)
+@patch("latigo.executor.PredictionExecutor._perform_auth_checks", new=MagicMock())
```
o-fedorov, line 12:
```suggestion
@patch("latigo.gordo.prediction_execution_provider.GordoClientPool")
@patch(
    "latigo.executor.model_info_provider_factory", side_effect=MagicMock(spec_set=GordoModelInfoProvider))
)
@patch("latigo.executor.PredictionExecutor._perform_auth_checks")
```
Default `new` argument is `MagicMock`. Also not sure what is the purpose of additional arguments to the patch of `"latigo.executor.model_info_provider_factory"`. Perhaps plain `@patch("latigo.executor.model_info_provider_factory", spec=GordoModelInfoProvider)` is enough.

o-fedorov: Also patching a private method looks wrong

tests/unit/test_executor.py
devProdigy: Thanks, but this won't work cause new object will be created later in the code.

tests/unit/test_executor.py
devProdigy: it is tested in scope of this test but that "private" method should be tested as well. Maybe it should not be private at all

tests/unit/test_executor.py
devProdigy: if you won't specify `new` then you should pass it as the argument and we don't need it in the function.
Plain won't work cause we're mocking a factory here.

Can't see nothing wrong in the private method mocking, sometimes like here you have to disable some inner logic and this is the way to do it. Why is it wrong from your perspective?

PR #76, opened 2020-05-19
17188 adjust error handling

devProdigy:
## Description
adjust error handling

* add how to run image locally to README.md;
* run black on few files;
* add custom exceptions;
* add param to enable/disable detailed executor logging;
* remove second container from docker-compose.

[Azure DevOps Link](https://dev.azure.com/EquinorASA/OMNIA%20Prevent/_workitems/edit/17188)

## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [X] Documentation content changes
- [X] Tests
- [ ] Other

# Checklist:


app/latigo/executor/__init__.py
```diff
@@ -48,9 +46,10 @@ def __init__(self, config: dict):
         self._prepare_executor()
         self._perform_auth_checks()
 
-    def _fail(self, message: str):
-        self.good_to_go = False
+    @staticmethod
+    def _fail(message: str):
         logger.error(message)
+        raise Exception(message)
```
o-fedorov, line 52: I would propose to have a more specific exception if this should be handled, alternatively use `sys.exit(message)`

app/latigo/executor/__init__.py
```diff
@@ -77,31 +76,23 @@ def _prepare_sensor_data_provider(self):
         self.sensor_data_provider_config = self.config.get("sensor_data", None)
         if not self.sensor_data_provider_config:
             self._fail("No sensor_data_config specified")
-        self.sensor_data_provider = sensor_data_provider_factory(
-            self.sensor_data_provider_config
-        )
+        self.sensor_data_provider = sensor_data_provider_factory(self.sensor_data_provider_config)
```
o-fedorov, line 79: Was it changed just due to automatic reformatting? If so, could you please revert the changes related to automatic formatting and apply them with another PR, to have less visual noise within current PR?

app/latigo/executor/__init__.py
```diff
@@ -292,118 +230,71 @@ def _get_tags_time_series_ids_for_model(self, prediction_data: PredictionDataSet
...
+        )
 
     @staticmethod
     def make_prediction_task_info(task: Task) -> str:
         """Make info about prediction task for logging."""
         return f"'{task.project_name}.{task.model_name}', prediction: from '{task.from_time}' to '{task.to_time}'"
 
     def _log_task_execution_time(
-            self, task: Task = None, label: str = "Total task execution time", chars_to_append: str = "\n\n"
+        self, label: str, task: Task = None, force_log_writing: bool = False, add_new_line: bool = False
```
o-fedorov, line 272: All this code related to logging should be extracted to a separate place, to not mix in business logic and logging facilities.

app/latigo/executor/__init__.py
```diff
@@ -133,132 +120,83 @@ def _perform_auth_checks(self):
...
-            self._log_task_execution_time(label="stored to Metadata API", chars_to_append="")
-
-        except Exception as e:
-            logger.error(
-                f"Could not store prediction or metadata for task {self.make_prediction_task_info(task)}. Error: {e}."
-            )
-            raise e
+            self._log_task_execution_time(label="stored to Metadata API")
+        except MetadataStoringError as e:
+            logger.error(self.format_error_message("Metadata was not stored", e=e, task=task))
```
o-fedorov, line 198: Why not keep it simple, like so:

```suggestion
            logger.exception("Metadata was not stored for task %s", task)
```

app/latigo/time_series_api/time_series_exceptions.py
```diff
@@ -0,0 +1,12 @@
+"""Exceptions for the Time Series logic."""
+from typing import List
+
+
+class NoCommonAssetFound(Exception):
```
o-fedorov, line 5: Instead of creating a custom exception, just use `LookupError` with properly formatted error message.

o-fedorov: Also see this section for more insights: https://github.com/amontalenti/elements-of-python-style#rarely-create-your-own-exception-types

app/latigo/metadata_api/metadata_exceptions.py
```diff
@@ -0,0 +1,9 @@
+"""Exceptions for the Metadata API logic."""
+
+
+class MetadataStoringError(Exception):
+    """Raise if got not 200 status code from the API."""
+
+    def __init__(self, message: str):
+        self.message = message
+        super().__init__(self.message)
```
o-fedorov, line 9:
```suggestion
class MetadataStoringError(Exception):
    """Raise if got not 200 status code from the API."""
```

The added functionality is redundant. Also consider raising existing exception instead of creating a new one.

app/latigo/executor/__init__.py
devProdigy: to keep particular logs the same structure (error should be logged as well). This is need for the following visualization on the dashboard and monitor alerts. And if we change anything it would be done in one place. Latigo struggles with unified logging thus debugging might be a nightmare.

app/latigo/metadata_api/metadata_exceptions.py
o-fedorov: Also could you please point where `MetadataStoringError` is raised?

app/latigo/executor/__init__.py
o-fedorov: There are better ways to support this:

1. Formatters: https://docs.python.org/3.7/library/logging.html#formatter-objects
2. Passing contextual information: https://docs.python.org/3/howto/logging-cookbook.html#adding-contextual-information-to-your-logging-output
3. Third-party libraries for additional context support, like this one: https://github.com/peopledoc/pylogctx

app/latigo/executor/__init__.py
devProdigy: this logic relies on the class property that freezes starting time before each task execution and to have control over it is good to have it here. Also this is related only to this class, so, I think, should stay here.

o-fedorov: There is plenty of ways for notifying logging facility about task starting without a need to integrate it's code inside business logic.

app/latigo/executor/__init__.py
devProdigy: yes, it's only cause of Black formatting.
I know, that it created some noice, but it'll take lots of time to fetch it from here and revert it, I'll do such next time I'll clean the code, sorry about that...

app/latigo/metadata_api/metadata_exceptions.py
devProdigy: Thanks for noticing, I missed to commit it, it should be [here](https://github.com/equinor/latigo/blob/master/app/latigo/metadata_api/metadata_storage_provider.py#L78(
The main idea why I used custom exceptions cause Latigo time to time raises some unexpected errors and to separate them from known one. And in-build Exceptions are too generic to catch them and ignore on some of the Dashboards.
I tend to leave custom Exceptions cause of the mentioned "surprises" it Latigo, can you offer smth else instead to solve such situation?

P.S. I'll remove the constructor.

app/latigo/time_series_api/time_series_exceptions.py
devProdigy: Thanks for the link, good point in general but I have doubts about using it in the current case cause of mentioned in my comment [here](https://github.com/equinor/latigo/pull/76#discussion_r427356715)

app/latigo/executor/__init__.py
devProdigy: done

app/latigo/metadata_api/metadata_exceptions.py
devProdigy: PR updated

app/latigo/executor/__init__.py
```diff
@@ -292,118 +227,71 @@ def _get_tags_time_series_ids_for_model(self, prediction_data: PredictionDataSet
...
+            - task: task to be additionally logged;
+            - force_log_writing: if True log will be written despite "self.log_debug_enabled";
+            - add_new_line: add `task_visual_separator` to the end of log.
+
+        Note: 'self.task_fetch_start' should be initialized previously.
+        """
+        if not self.log_debug_enabled and not force_log_writing:  # TODO Alex test this
+            return
+
+        task_visual_separator = "\n\n"
```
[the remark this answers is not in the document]
devProdigy: done

app/latigo/executor/__init__.py
```diff
@@ -1,42 +1,42 @@
...
+from latigo.time_series_api.time_series_exceptions import NoCommonAssetFound
+from latigo.types import PredictionDataSet, SensorDataSpec, Task
+from latigo.utils import human_delta
 
-EXCEPTIONS_TO_HANDLE = (Exception, ResourceGone, NotFound, BadGordoRequest)
+GORDO_EXCEPTIONS = (ResourceGone, NotFound, BadGordoRequest, HttpUnprocessableEntity)
+GORDO_ERROR_IDENTIFIER = "Gordo error"
+IOC_DATA_EXCEPTIONS = (InsufficientDataAfterRowFilteringError, InsufficientDataError, NoTagDataInDataLake)
+IOC_ERROR_IDENTIFIER = "Data error"
+TASK_LOG_VISUAL_SEPARATOR = "\n\n"
```
o-fedorov, line 32: It should not be a constant. It even does not deserve a variable. Just use this string in the text directly. The same is true for `GORDO_ERROR_IDENTIFIER` and `IOC_ERROR_IDENTIFIER`.

The case with report generation is that better readability & maintainability means that all text of a report should be placed in relatively the same place (high cohesion).

app/latigo/metadata_api/metadata_exceptions.py
```diff
@@ -0,0 +1,7 @@
+"""Exceptions for the Metadata API logic."""
+
+
+class MetadataStoringError(Exception):
+    """Raise if got not 200 status code from the API."""
+
+    pass
```
o-fedorov, line 7:
```suggestion
class MetadataStoringError(Exception):
    """Raise if got not 200 status code from the API."""
```

`pass` is redundant since there is already a docstring in the class body.

app/latigo/metadata_api/metadata_storage_provider.py
```diff
@@ -75,7 +76,7 @@ def put_prediction_metadata(
         )
         res = self.send_time_series_id_metadata(ts_metadata)
         if res.status_code != 200:
```
o-fedorov, line 78: I suppose it's a `requests` response. If so, just do `res.raise_for_status()`, and then catch the `HTTPError` in the upstream code.

app/latigo/executor/__init__.py
devProdigy: @zveres-ss asked to give him ability to reuse `TASK_LOG_VISUAL_SEPARATOR` in the future, please deal between you two about it :)

@o-fedorov  GORDO_ERROR_IDENTIFIER and IOC_ERROR_IDENTIFIER are reused in tests, so it gas some value to use it here but not crucial, I can move it from here, if it's is the right thing to do.

o-fedorov:
My main recomendations remain the same: decouple logging functionality from the rest of the code.

1. Extract custom logging code to a dedicated module or package.
2. Use [logging.Formatter](https://docs.python.org/3.7/library/logging.html#formatter-objects) subclasses to produce logs format suitable for parsing by the logs aggregation tool.
3. For passing logging context consider using [pylogctx](https://github.com/peopledoc/pylogctx). It works greatly for synchronous code.

Also note that integrating logging functionality to deeply inside the business logic means that it brings extra dependency with the logging dashboard.  That is, for any change to business logic one should check if the dashboard still works as expected.  And chances are that someday it will be broken.

When logging and the rest of the logic is decoupled, one can simply use `logger.info(...)`, `logger.exception(...)` and so on, and the rest is to be handled by the logging machinery in a consistent way.

app/latigo/metadata_api/metadata_storage_provider.py
devProdigy: good idea, done

app/latigo/metadata_api/metadata_exceptions.py
devProdigy: deleted, this exception is not needed anymore

devProdigy:
> My main recomendations remain the same: decouple logging functionality from the rest of the code.
> 
> 1. Extract custom logging code to a dedicated module or package.
> 2. Use [logging.Formatter](https://docs.python.org/3.7/library/logging.html#formatter-objects) subclasses to produce logs format suitable for parsing by the logs aggregation tool.
> 3. For passing logging context consider using [pylogctx](https://github.com/peopledoc/pylogctx). It works greatly for synchronous code.
> 
> Also note that integrating logging functionality to deeply inside the business logic means that it brings extra dependency with the logging dashboard. That is, for any change to business logic one should check if the dashboard still works as expected. And chances are that someday it will be broken.
> 
> When logging and the rest of the logic is decoupled, one can simply use `logger.info(...)`, `logger.exception(...)` and so on, and the rest is to be handled by the logging machinery in a consistent way.

It might be a good recommendation if we know how logging should be, what dashboards should be, will dashboards sustain or gone away in few weeks, etc, but we don't have any requirements for this. For now it seems for me as overwork for such task current PR is related to with the time, that we have for it. Thus this won't be implemented now and in the next sprint until we see the "big picture" and what are the real needs.
P.S. other comments were taken into the account and resoled.

devProdigy:
closed by accident

app/latigo/time_series_api/time_series_exceptions.py
o-fedorov: 1. Could you either define a single module for a custom exception, or define it in the same module where it is raised? That improves the code cohesion.
2. Consider have less logic inside the exception.  That gives more flexibility   For example, just define the exception like this:
```
class NoCommonAssetFound(Exception):
    """Raise if no 'common asset' found in dateframe column headers."""
```

And then use it as following
```
parsed_data = '; '.join(col[1] for col in column_values)
logger.error("No common asset found in dataframe columns.  Parsed data: %s", parsed_data})
raise NoCommonAssetFound("No common asset found in dataframe columns.")
```

That gives better flexibility on the information logged out, and keeps the information in the place where it is defined.  That is, high-level `executor` code knows nothing about `parsed_data` at the place where `NoCommonAssetFound` is handled.

o-fedorov:
I insist on decoupling logging from business logic now.

Just one example: `format_error_message` method can be completely replaced with a Formatter subclass, or even a Filter-Formatter chain.

If we do not decouple logging functionality now when the changes are introduced, it will be much harder to do so in the future.

devProdigy:
> I insist on decoupling logging from business logic now.
> 
> Just one example: `format_error_message` method can be completely replaced with a Formatter subclass, or even a Filter-Formatter chain.
> 
> If we do not decouple logging functionality now when the changes are introduced, it will be much harder to do so in the future.

@o-fedorov 
I'm not aware how to do it in the same understandable way as I have with what you suggest for all following points:
1. we should have different formats for error message, ordinary logging and other logging (with supporting the SAME `if else` logic, as well as `self.log_debug_enabled and not force_log_writing` logic).
2. firstly init and use `self.task_fetch_start` for `_log_task_execution_time` without passing it to the function each time and "dark magic".

You have to help me with this then, otherwise this PR will be left unchanged. 
I don't see the point just to move one function to another file and then struggle with initiating start execution time and other stuff, thus probably making code more complex. As for me this is used/related only in this module, does not touch the business logic and you still have to call that methods from the same parts. So it's up to you: you can adjust this PR with commits that will do the "stuff" with easy and understandable way or approve the PR cause I can't understand you point of view here.

o-fedorov:
> You have to help me with this then, otherwise this PR will be left unchanged.

If you wish I could provide related code changes to Latigo. Letting the PR for your own judgements.

PR #77, opened 2020-05-20
18655 Remove milliseconds for task timeframes, remove redundant utils.

o-fedorov:
This fixes errors like this one:

```
  File "/app/.local/lib/python3.7/site-packages/gordo/machine/dataset/base.py", line 138, in join_timeseries
    aggregation_methods=aggregation_methods,
  File "/app/.local/lib/python3.7/site-packages/gordo/machine/dataset/base.py", line 198, in _resample
    raise RuntimeError(msg)
RuntimeError: Error - for TRA-35TT8566.PV, first timestamp 2020-05-20 08:10:01+00:00 is before the resampling start point 2020-
05-20 08:10:01.382657+00:00
```

Also fix some "broken windows".

## Description

Summary of the changes / which issue is fixed / etc.

[Azure DevOps Link](https://dev.azure.com/EquinorASA/OMNIA%20Prevent/_sprints/taskboard/OMNIA%20Prevent%20Team/OMNIA%20Prevent/Sprint%209?workitem=18655)

## Type of change

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

# Checklist:


app/latigo/scheduler/__init__.py
```diff
@@ -158,10 +162,13 @@ def perform_prediction_step(self):
         stats_models_ok = {}
         stats_projects_bad = {}
         stats_models_bad = {}
-        stats_start_time_utc = get_datetime_now_in_utc()
-        # 'prediction_start_time' and 'prediction_end_time' should be in UTC time cause Queue will ignore timezone.
-        prediction_start_time_utc = stats_start_time_utc - self.continuous_prediction_delay
-        prediction_end_time_utc = prediction_start_time_utc + self.continuous_prediction_interval
+
+        stats_start_time_utc = datetime.datetime.now(utc)
```
devProdigy, line 166: what is the reason to replace `get_datetime_now_in_utc()` with  `datetime.datetime.now(utc)`? It does the same but dev won't forget to use timezone.
It can be reused in other places and people will be aware that such function exists and this means that UTC time should be used.

o-fedorov: If you take a look at the module code you’ll see that `datetime.now()` is already used in multiple places. Having a "utility" function for one-line command just decrease code cohesion without bringing any value.

Also I’m not afraid that one would forget about using `utc` for the variable with `utc` suffix, covered with tests and having a comment about using `uts` timezone.

devProdigy: The value is that you don't have to import `pytz` and add it to the datetime each time you need `now`. And the code should be refactored to be in the same timezone. When we'll start doing that we'll need that function and no "code cohesion" as for me here.

o-fedorov: Don’t see a problem with the import. This is a common way to get current timezone-aware datetime. If you wish I could use `datetime.timezone.utc` instead.

o-fedorov: This function does not save time, does not save place and does not improve readability.

And if I needed to get a timezone-aware datetime, I even will not think of looking for something like this, but instead would go for the common approach.

What’s more important, is that if you need to know what the function does, you need to go to it’s definition and check. And you *should* do it, because one may place `datetime.utcnow()` instead of `datetime.now(utc)`.

devProdigy: Still as for such point of view is very individually, but let it be, there's usually more than one way to do the same thing.

PR #84, opened 2020-05-26
Feature/use redis service as as cache

devProdigy:
## Description
use redis service as as cache

Changes to infrastructure was already [merged](https://github.com/equinor/latigo-k8s/pull/26)

additionally:
* adjust default values in configs;
* adjust readme;
* remove unused imports.

[Azure DevOps Link](https://dev.azure.com/EquinorASA/OMNIA%20Prevent/_workitems/edit/17812)

## Type of change

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

# Checklist:


app/latigo/time_series_api/client.py
```diff
@@ -1,30 +1,12 @@
...
-from latigo.sensor_data import SensorDataProviderInterface
-from latigo.prediction_storage import PredictionStorageProviderInterface
-from latigo.utils import rfc3339_from_datetime
-
-from .misc import _itemes_present, _get_auth_session, _parse_request_json
-from .cache import *
+from latigo.types import TimeRange
+
+from .cache import TagMetadataCache
+from .misc import _get_auth_session, _itemes_present, _parse_request_json
```
o-fedorov, line 9: Private entities should not be used outside of their module.

app/latigo/time_series_api/client.py
```diff
@@ -57,38 +38,25 @@ def _parse_auth_config(self):
...
-        if not self.ims_subscription:
-            return self._fail("Could not create ims_subscription")
-
     def __init__(self, config: dict):
         self.good_to_go = True
         self.config = config
         if not self.config:
             raise Exception("No config specified")
-        self.meta_data_cache = MetaDataCache()
+        self.tag_metadata_cache = TagMetadataCache(
```
o-fedorov, line 46: [Dependency inversion principle](https://en.wikipedia.org/wiki/Dependency_inversion_principle) states that high-level code should be independent of the low-level implementation details.  That means that `TagMetadataCache` should be configured outside of `TimeSeriesAPIClient`.

`TimeSeriesAPIClient` should not know about how `TagMetadataCache` is configured, it should only utilize the client interface.

app/latigo/utils.py
```diff
@@ -173,6 +173,18 @@ def load_configs(
     return config, None
 
 
+def update_config_with_cache_accesses(config: dict):
+    """Populate config with cache credentials."""
+    keys_to_populate = ["prediction_storage", "sensor_data"]
```
o-fedorov, line 178: It's better to use it vice-versa, that is, use purely environment variables to configure the client.  It's a part of [12-factor app](https://12factor.net/config) approach, that covers the needs of kubernetes applications the most.

Latigo config is already [complicated enough](https://en.wikipedia.org/wiki/God_object), and this method adds another implicit modification to the configuration object.

app/latigo/time_series_api/client.py
devProdigy: Agree, such refactoring should be made eventually, but I only removed unused and auto-regroup imports in this PR, refactoring "private" methods and maybe what is inside them will be too out of scope of this PR.

app/latigo/time_series_api/client.py
devProdigy: To clarify: you offer to create constructor for `TagMetadataCache` and pass `config` inside it and there unpack needed values? I'll use ENV variables instead but just to understand the point.

app/latigo/utils.py
devProdigy: Agree, I'll use env variables.

app/latigo/time_series_api/client.py
o-fedorov: Oh, did not notice. Good then.

app/latigo/time_series_api/client.py
o-fedorov: The simplest way is to place it inside boilerplate code section. That is, initialize `TagMetadataCache` in `latigo/bin/executor.py` under `if __name__ == "__main__":` section and then pass it to `TimeSeriesAPIClient` constructor somehow.

o-fedorov: Or even initialize `TimeSeriesAPIClient` in the boilerplate code and pass it to Executor.

devProdigy: Then we have to pass it trough all the logic-chain that we have and it'll be too many of such objects as a params cause Executor uses lots of different objects that similar with this one.
I'll stick with the approach with constructor not to make it too complicated.

devProdigy: done

app/latigo/utils.py
devProdigy: done

app/latigo/time_series_api/client.py
o-fedorov: In ideal world it would be like this:

1. Initialize the cache client.
2. Initialize the timeseries client with the cache client.
3. Initialize Executor with the timeseries client.

o-fedorov: Also there are tools that greatly simplify the boilerplate code, my favorite one is this tiny lib: https://github.com/ivankorobkov/python-inject

devProdigy: Can't evaluate your idea fully, but current architecture has lot's of dependencies/inheritance of the `TimeSeriesAPIClient` and I don't think that this is a good time for such improvements.
Let's discuss architecture refactoring when we came closer its implementation.

app/bin/executor.py
```diff
@@ -31,6 +31,8 @@
         logger.error(f"Could not load configuration for executor: {err}")
         sys.exit(1)
 
+    update_config_with_cache_accesses(config)
```
o-fedorov: Is it still needed?

devProdigy: no, it was deleted, please try to refresh the page.

tests/unit/test_time_series_api.py
```diff
@@ -124,18 +124,3 @@ def test_transform_from_timeseries_to_gordo():
     # logger.info(pprint.pformat(gordo_data))
     expected_gordo_data = {"X": [[69.69, 42], [42.69, 420], [1337.69, 420]]}
     assert gordo_data == expected_gordo_data
-
```
o-fedorov, line 127: It would be nice to have some tests.

app/latigo/time_series_api/cache.py
```diff
@@ -1,53 +1,51 @@
...
-        name: str,
-        asset_id: typing.Optional[str],
-        meta: typing.Optional[typing.Dict],
-    ):
-        key = self._key(name, asset_id)
-        if not meta:
-            del self._data[key]
-        else:
-            self._data[key] = meta
+class TagMetadataCache(StrictRedis):
```
o-fedorov, line 11: Using delegation instead of inheritance would simplify testing.  That is, if you have code like this:

```
class TagMetadataCache:
    def __init__(self, ...):
        self._cache = StrictRedis(...)


    def get_meta(self, ...):
        res = self._cache.get(key)
```

then you just need to mock `StrictRedis` to verify all the code in this module.

o-fedorov: Also delegation is more semantically valid if `TagMetadataCache` is not supposed to be a kind of `StrictRedis` (that is, if there is no need to use `TagMetadataCache` and `StrictRedis` interchangeably).

o-fedorov:
I checked Latigo code and, yes, there is no easy way to provide components boilerplate directly.

Using environment variables in a class constructor looks like the best of a bad lot, still the configuration is hidden inside the code and that may be surprising for future maintainers.

Still, with a help of third-party libraries it is possible to decouple interfaces → from implementation → from configuration.

For example, if you utilize [python-inject](https://pypi.org/project/Inject/), the changes may be as following:

```
# in app/bin/executor.py
import inject


def inject_config(binder):
    """Application components boilerplate."""
    binder.bind(TagMetadataCache, TagMetadataCache(<connection params>))

if __name__ == "__main__"
    # Configure all dependencies only when the service is ready
    inject.configure(inject_config, bind_in_runtime=False)


# In  app/latigo/time_series_api/client.py
class TimeSeriesAPIClient:
    _tag_metadata_cache = inject.attr(TagMetadataCache)


# In tests/conftest.py
@pytest.fixture(autouse=True)
def configure_dependencies():
    """Configure application components."""
    inject.clear_and_configure(lambda binder: binder.bind(TagMetadataCache, Mock(spec=TagMetadataCache), bind_in_runtime=False)
    yield
    inject.clear()
```

app/latigo/time_series_api/client.py
```diff
@@ -57,38 +38,20 @@ def _parse_auth_config(self):
...
-        if not self.ims_subscription:
-            return self._fail("Could not create ims_subscription")
-
     def __init__(self, config: dict):
         self.good_to_go = True
         self.config = config
         if not self.config:
             raise Exception("No config specified")
-        self.meta_data_cache = MetaDataCache()
+        self.tag_metadata_cache = TagMetadataCache()
```
o-fedorov, line 46: It's better to make the method private.

app/latigo/time_series_api/client.py
```diff
@@ -145,15 +108,15 @@ def _get_meta_by_name_raw(
         return _parse_request_json(res)
 
     def _get_meta_by_name(
-        self, name: str, asset_id: typing.Optional[str] = None
+        self, name: str, asset_id: str
     ) -> typing.Tuple[typing.Optional[typing.Dict], typing.Optional[str]]:
-        if self.meta_data_cache:
-            meta = self.meta_data_cache.get_meta(name, asset_id)
+        if self.tag_metadata_cache:
```
o-fedorov, line 113: Is there a reason why `self.tag_metadata_cache` may be `False`?

devProdigy: haven't looked at it, some code adjustments will be done in the next task of fixing outdated caching.

devProdigy:
> inject.attr(TagMetadataCache)

@o-fedorov  interesting approach, but when I tried it got:
`AttributeError: '_AttributeInjection' object has no attribute 'get_metadata'`, is something is missing? Seems like `inject.attr(TagMetadataCache)` does not do its job.

o-fedorov:
> AttributeError: '_AttributeInjection' object has no attribute 'get_metadata', is something is missing? Seems like inject.attr(TagMetadataCache) does not do its job.

Do you use it in class-level or instance level? There are multiple ways to retrieve an injected instance, for example:

```
# Assign to a class attribute with `.attr` method:
class TimeSeriesAPIClient:
    _tag_metadata_cache = inject.attr(TagMetadataCache)

# Use inside code with `.instance` method:
class TimeSeriesAPIClient:
    def __init__(self, ...):
        self._tag_metadata_cache = inject.instance(TagMetadataCache)
```

devProdigy:
> > AttributeError: '_AttributeInjection' object has no attribute 'get_metadata', is something is missing? Seems like inject.attr(TagMetadataCache) does not do its job.
> 
> Do you use it in class-level or instance level? There are multiple ways to retrieve an injected instance, for example:
> 
> ```
> # Assign to a class attribute with `.attr` method:
> class TimeSeriesAPIClient:
>     _tag_metadata_cache = inject.attr(TagMetadataCache)
> 
> # Use inside code with `.instance` method:
> class TimeSeriesAPIClient:
>     def __init__(self, ...):
>         self._tag_metadata_cache = inject.instance(TagMetadataCache)
> ```

I see, my bad, used on the instance level.
Nice :)

app/latigo/time_series_api/cache.py
devProdigy: done

app/latigo/time_series_api/client.py
devProdigy: done

tests/unit/test_time_series_api.py
devProdigy: Will do with next commit

devProdigy: done

app/latigo/time_series_api/client.py
```diff
@@ -1,30 +1,14 @@
...
-import datetime
-import json
-import pprint
-from requests.exceptions import HTTPError
-import pandas as pd
-import urllib.parse
+
+import typing
+
+import inject
```
o-fedorov, line 5: Is it still used here?

app/latigo/time_series_api/client.py
```diff
@@ -1,30 +1,14 @@
...
-import math
-import requests
-import datetime
-import json
-import pprint
-from requests.exceptions import HTTPError
-import pandas as pd
-import urllib.parse
+
+import typing
```
o-fedorov, line 3:
```suggestion
import typing
```
Just a nit-pick.

app/latigo/time_series_api/misc.py
```diff
@@ -53,7 +53,7 @@ def _itemes_present(res: dict) -> bool:
     return True
 
 
-def _get_auth_session(auth_config: dict, force: bool = False):
+def get_auth_session(auth_config: dict, force: bool = False):
```
o-fedorov, line 56: I would make a separate PR for a change like that next time.

app/latigo/time_series_api/client.py
devProdigy: Thanks, adjusted.
for such things we have to enable back and isort for all the project in the future, sadly now we don't have time for such cleaning...

PR #89, opened 2020-06-01
Feature/18969 use multiple assets from metadata api

devProdigy:
## Description

Feature to use multiple assets/projects from Metadata API to process all existing configs and do not paste them manual.
Note: we have to clean our DBs on the envs to have only projects that are in Gordo, otherwise app will fall. For this code is catching such error and logging the error for now.

[Azure DevOps Link](https://dev.azure.com/EquinorASA/OMNIA%20Prevent/_workitems/edit/18969)

**TODOs before apply changes on the ENVs:** 
* clean the configs (executor and scheduler) on DEV and PROD from "projects"
* update config with “models_metadata_info”

## Type of change

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] CI related changes
- [X] Tests
- [ ] Other

# Checklist:


app/latigo/gordo/model_info_provider.py
```diff
@@ -30,22 +29,30 @@ def __str__(self):
...
+            raise ValueError("'projects' can not be empty.")
         for project_name in projects:
-            # logger.info(f"LOOKING AT PROJECT {project_name}")
             client = self.gordo_pool.allocate_instance(project_name)
             if client:
-                machines += client._get_machines(machine_names=model_names)
+                try:
+                    machines += client._get_machines(machine_names=model_names)
+                except TypeError as e:
+                    if "byte indices must be integers or slices, not str" not in str(e):
```
devProdigy, line 43: weird error, but this is what Gordo client is throwing in our case.
See PR description for explaining this case.

app/latigo/model_metadata_info/__init__.py
```diff
@@ -0,0 +1,26 @@
+import logging
+from typing import List
+
+from latigo.metadata_api.client import MetadataAPIClient
+
+logger = logging.getLogger(__name__)
+
+
+class ModelMetadataInfoInterface:
```
devProdigy, line 9: I don't like how architecture was made with this interfaces and other stuff, but due to kipping all same and not to waste lots of time of smth that is not related to the PR - I left everything in one file in such way.
If we be lucky and have time/resources for refactoring this and similar places should be taken into work.

app/latigo/scheduler/__init__.py
```diff
@@ -57,9 +59,20 @@ def _prepare_task_queue(self):
         if not self.task_queue:
             self._fail("No task queue configured")
 
+    def _prepare_models_metadata_info_provider(self):
```
devProdigy, line 62: same comment as for Interface itself, this piece of code follows the approach for all other "modules".

devProdigy:
If this PR will be fast enough I'll try to remove broken and unused tests to run all folder with coverage :)
That's the dream!

app/bin/scheduler.py
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 import logging
 
-from common import basic_config
+from bin.common import basic_config
```
o-fedorov, line 4: Could you please doublecheck if this change works in docker/kubernetes, since AFAIK the directory containing `bin` is not in PYTHONPATH there.

app/latigo/gordo/client_pool.py
```diff
@@ -120,16 +119,14 @@ def allocate_instance(self, project: str):
...
-        projects = self.config.get("projects", [])
-        if not isinstance(projects, list):
-            projects = [projects]
-        for project in projects:
-            self.allocate_instance(project)
+    def delete_instance(self, project_name: str):
+        """Delete Gordo client instance from pool."""
+        client = self.client_instances_by_project.get(project_name, None)
+        if client:
+            self.client_instances_by_project.pop(project_name)
```
o-fedorov, line 126:
```suggestion
        self.client_instances_by_project.get(project_name, None)
```

app/latigo/gordo/model_info_provider.py
```diff
@@ -30,22 +29,30 @@ def __str__(self):
...
-            if not isinstance(projects, list):
-                projects = [projects]
+            raise ValueError("'projects' can not be empty.")
         for project_name in projects:
-            # logger.info(f"LOOKING AT PROJECT {project_name}")
             client = self.gordo_pool.allocate_instance(project_name)
             if client:
-                machines += client._get_machines(machine_names=model_names)
+                try:
+                    machines += client._get_machines(machine_names=model_names)
```
o-fedorov, line 41: You should not call the private method.

app/latigo/gordo/model_info_provider.py
```diff
@@ -30,22 +29,30 @@ def __str__(self):
...
+                except TypeError as e:
+                    if "byte indices must be integers or slices, not str" not in str(e):
+                        raise
+
+                    # this is case when "project_name" is invalid.
+                    # we will skip such project cause API might be not in sync with Gordo.
+                    logger.error(f"{e}", stack_info=True, exc_info=True)
```
o-fedorov, line 48:
```suggestion
                    logger.exception("Invalid project: %s", project_name)
```

o-fedorov: 1. Use `logger.exception` to print out the exceptions.
2. Arguments substitution (with `%s` or `%r`) is a preferred way for log messages parametrization, as it provides extra data to the logs formatter.

app/bin/scheduler.py
devProdigy: It works, I did the same for executor in previous PR and checked before merging 
https://github.com/equinor/latigo/blob/master/app/bin/executor.py#L8

But I've rechecked and it works, thanks for keeping eye on that.

app/latigo/gordo/client_pool.py
devProdigy: Sorry, didn't get the idea of the suggestion. What did you mean?

app/latigo/gordo/model_info_provider.py
devProdigy: This is the only way to grab needed data from the Gordo client. It's too much to refactor the client for this perspective, cause this method should be public.

app/latigo/model_metadata_info/__init__.py
o-fedorov: That's what we call "broken windows".  I would propose to take a look at a possibility to replace this approach with `inject`, because everything this module does is either provides `MetadataAPIClient` or raises `ValueError` if one can't be configured.

Here is an example how it could be done:

```
from functools import partial


def inject_config(binder, *, config):
    ...
    models_metadata_info = config["models_metadata_info"]
    binder.bind(MetadataAPIClient, MetadataAPIClient(models_metadata_info))


inject.configure(partial(inject_config, config=config), bind_in_runtime=False)
```

app/latigo/gordo/model_info_provider.py
devProdigy: done

app/latigo/scheduler/__init__.py
o-fedorov: The comment is the same as above.  Also I would move the auth check to the boilerplate code, as it is more related to the configuration, not to the class itself.

app/latigo/scheduler/__init__.py
```diff
@@ -162,16 +166,11 @@ def perform_prediction_step(self):
         for model in models:
             project_name = model.project_name
             if not project_name:
-                logger.warning("No project name found for model, skipping model")
-                continue
+                raise ValueError(f"No project name found for model.")
```
o-fedorov, line 169: Could you please explain why `ValueError` is raised here? I would expect that if there is no project name or model name for particular model, we still should continue with the next models.

app/latigo/model_metadata_info/__init__.py
devProdigy: Agree that approach should be discussed and change but for hole system, not for one part.
Sorry. but I don't think that this PR is the right place for such architecture and tests refactoring, cause part of the code and architecture will have one approach and second one another.

app/latigo/scheduler/__init__.py
```diff
@@ -104,24 +117,15 @@ def _prepare_scheduler(self):
...
-                self._fail(f"Error in project spcification: {p}")
-        except Exception as e:
-            self._fail(f"Could not parse '{p}' into projects: {e}")
         self.run_at_once = self.scheduler_config.get("run_at_once", True)
         self.back_fill_max_interval = pd.to_timedelta(
             self.scheduler_config.get("back_fill_max_interval", "1d")
         )
+
+        # fetch projects from the API
+        self.projects = self.models_metadata_info_provider.get_projects()
```
o-fedorov, line 126: It's better to fetch the projects in `perform_prediction_step`, that is fetch the projects right before processing.

app/latigo/scheduler/__init__.py
devProdigy: Same answer as on the comment above
https://github.com/equinor/latigo/pull/89#discussion_r433686880

app/latigo/scheduler/__init__.py
devProdigy: cause in this case something very unusual happened and we have to take care about now, without loosing it in the logs.
Such case is mostly unlikely, but still.

app/latigo/gordo/client_pool.py
o-fedorov: Sorry, there should've been `pop` instead of `get`, updated.

devProdigy: I see, good point, thanks.
Done!

app/latigo/scheduler/__init__.py
o-fedorov: Then increase the severity of the log and/or configure an alert.

Still one broken model should not prevent processing of other good models.

app/latigo/scheduler/__init__.py
devProdigy: Thanks, didn't notice that they won't be reloaded each time.
Code was changed.

app/latigo/scheduler/__init__.py
devProdigy: Won't argue, done.

tests/conftest.py
```diff
@@ -64,6 +67,11 @@ def schedule_config(auth_config):
             "type": "kafka",
             "connection_string": "Endpoint=sb://sb/;SharedAccessKeyName=name;SharedAccessKey=yd;EntityPath=path",
             "topic": "latigo_topic",
+        },
+        "models_metadata_info": {
```
o-fedorov, line 71: Why not just have one single client for MetadataAPI, and one single configuration for it?

tests/unit/gordo_components/test_gordo_model_info_provider.py
```diff
@@ -0,0 +1,23 @@
+from unittest.mock import Mock, patch
+
+import pytest
+
+from tests.unit.gordo_components.conftest import parametrized_get_model_data
+
+
+@pytest.mark.parametrize("data", parametrized_get_model_data)
```
o-fedorov, line 8:
```suggestion
@pytest.mark.parametrize("projects", [["project-1", "project-2"], ["project-1", "missing_project", "project-2"]])
```
and update the tests accordingly.  Also get rid of `parametrized_get_model_data`.  Having a globally available object `parametrized_get_model_data` with lists, introduces a shared state which may lead to hard-to-debug issues if one of the tests alters a list.

Also instead of parametrization it seems it would be better to have 2 separate tests with models and side effects provided for each project explicitly.

tests/unit/gordo_components/test_gordo_model_info_provider.py
```diff
@@ -0,0 +1,23 @@
...
+    expected_models = data.get("models")
+    models_side_effects = []
+
+    for project in projects:
+        if project == "missing_project":
+            models_side_effects.append(TypeError("byte indices must be integers or slices, not str"))
+        else:
+            models_side_effects.append([m for m in expected_models if m.project_name == project])
+
+    with patch("latigo.gordo.client_pool.Client._get_machines", new=Mock(side_effect=models_side_effects)):
```
o-fedorov, line 21:
```suggestion
    with patch("latigo.gordo.client_pool.Client._get_machines", side_effect=models_side_effects):
```

tests/conftest.py
devProdigy: this is configuration for scheduler, not for executor.
If you mean merge configs and make one common and maybe separate for executor and scheduler - it's out of this PR scope.

tests/unit/gordo_components/test_gordo_model_info_provider.py
devProdigy: done

tests/unit/gordo_components/test_gordo_model_info_provider.py
devProdigy: `parametrized_get_model_data` in the global scope make sense - moved to test.
But I don't see the reason of splitting the test on two separate, the code will be almost the same.

PR #90, opened 2020-06-03
19231 Fix executor freezing on KeyboardInterrupt.

o-fedorov:
* Get rid of the poll timeout as it tends to block executor.
* Explicitly close the consumer on executor exiting.

[Azure DevOps Link](https://dev.azure.com/EquinorASA/OMNIA%20Prevent/_workitems/edit/19231)

## Type of change

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

### DIFF
--- a/README.md
+++ b/README.md
@@ -305,7 +305,6 @@ Both scheduler and executor has a task_queue configuration. It describes their c
 |      ---: | :-----: | :---------- |
 | type | "kafka" | Currently only kafka is supported. There has been "event_hub" and "postgres" in the past. |
 | connection_string | not shown | The full connection string as copied from event_hub panel in azure portal. |
-| poll_timeout_sec | 100000 | Kafka timeout when waiting for response in client. |
 | security.protocol | "SASL_SSL" | See [confluent docs](https://docs.confluent.io/current/installation/configuration/consumer-configs.html). |
 | ssl.ca.location | "/etc/ssl/certs/ca-certificates.crt" | See [confluent docs](https://docs.confluent.io/current/installation/configuration/consumer-configs.html). |
 | sasl.mechanism | "PLAIN" | See [confluent docs](https://docs.confluent.io/current/installation/configuration/consumer-configs.html). |
--- a/app/latigo/task_queue/kafka.py
+++ b/app/latigo/task_queue/kafka.py
@@ -66,7 +66,6 @@ class KafkaTaskQueueSender(TaskQueueSenderInterface):
         # Producer configuration
         # See https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md
         # See https://github.com/edenhill/librdkafka/wiki/Using-SSL-with-librdkafka#prerequisites for SSL issues
-        self.poll_timeout_sec = config.get("poll_timeout_sec", 100)
         self.config, self.topic, err = prepare_kafka_config(config)
         if not self.config:
             raise Exception(f"No config parsed: {err}")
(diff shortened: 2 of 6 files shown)


app/latigo/task_queue/kafka.py
```diff
@@ -217,21 +223,17 @@ def _receive_event(self, timeout=5) -> typing.Optional[bytes]:
...
+    def _receive_event_with_backoff(self, backoff_sec=60) -> typing.Optional[Task]:
         """Poll queue for the Message with passed timeout.
 
         Args:
-            timeout_sec: Maximum time to block waiting from queue for message, event or callback (seconds);
             backoff_sec: Time for Thread to sleep before next try to receive the Message from queue.
         """
         task: typing.Optional[Task] = None
-        task_bytes = self._receive_event(timeout=timeout_sec)
+        task_bytes = self._receive_event()
```
devProdigy, line 233: what behaviour will we see if no task be in the queue on the moment of sych request?

README.md
```diff
@@ -305,7 +305,6 @@ Both scheduler and executor has a task_queue configuration. It describes their c
 |      ---: | :-----: | :---------- |
 | type | "kafka" | Currently only kafka is supported. There has been "event_hub" and "postgres" in the past. |
 | connection_string | not shown | The full connection string as copied from event_hub panel in azure portal. |
-| poll_timeout_sec | 100000 | Kafka timeout when waiting for response in client. |
```
devProdigy, line 308: please, remove this value from the configs from PROD and DEV before deploying to envs.

o-fedorov: Sure.

- [ ] remove this value from the configs from  DEV
- [ ] remove this value from the configs from  PROD

app/latigo/task_queue/kafka.py
devProdigy: asking cause I can't remember what errors/warnings/etc. we will get if let's say we don't have any messages to fetch for lets say 20 minutes (like on DEV). Will our subscription still be valid? Could you please check it?

o-fedorov: Executor prints out `No bytes in the message after polling. Sleeping for backoff - 60 seconds.`, then sleeps for 60 seconds and starts over.

That is to say, the pull is executed once per minute with at most 1 second timeout.

o-fedorov: > say we don't have any messages to fetch for lets say 20 minutes (like on DEV). Will our subscription still be valid? Could you please check it?

Sorry, I don't understand what you mean.  Do you expect that there is some code in Latigo that handles such cases in some special way?  Or some alert in Azure Monitor?

o-fedorov: Also I don't see no specific behavior for this case.  BTW, the code from the official documentation even does not have a backoff timeout, so the queue is pulled every second, and in case there is no data the loop simply starts over.

devProdigy: not about Azure Monitor.
Comment was totally related to the subscription behaviour if no messages were pulled for long period of time. Faced some issues maybe related to it when just started working with Latigo.

If you have a chance, could you please:
- run Latigo on the queue with one message, it'll consume it;
- then leave Latigo working for 25 minutes;
- check the logs for an errors;
- after produce one message and check that this message was consumed by queue without any errors or smth like this.

o-fedorov: No unusual behavior was observed.

o-fedorov:
UPD: added graceful shutdown on keyboard interrupt here: https://github.com/equinor/latigo/pull/90/files#diff-ab28f04a938ce4fa424e4aee005621ddR227

app/latigo/task_queue/kafka.py
devProdigy: thanks for checking.

PR #94, opened 2020-06-05
remove unused IMS files and refactor one function

devProdigy:
## Description

remove unused IMS files and refactor one function

## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

# Checklist:


app/latigo/time_series_api/client.py
```diff
@@ -88,26 +88,24 @@ def _post(self, *args, **kwargs):
...
+        url = f"{self.base_url}/{tag_id}/data"
         params = {
             "startTime": time_range.rfc3339_from(),
             "endTime": time_range.rfc3339_to(),
             "limit": 100000,
             "includeOutsidePoints": False,
         }
         res = self._get(url=url, params=params)
-        return _parse_request_json(res)
+        return parse_request_json(res)
```
o-fedorov, line 106: I see that it's just renaming the function and changing its signature, still there is a room for improvements. Using a function like this looks like applying DRY principle, but actually it decreases the code _cohesion_.

Really good DRY change would be to call `res.raise_for_status(); return res.json()` inside `self._get` and `self._post`, also do it explicitly, since there might be cases when additional code should be inserted after `res.raise_for_status()`, there may be content other than `json` and so on.

devProdigy: I see the point to call `res.raise_for_status(); return res.json()` inside `self._get` and `self._post`, additionally `self._get` and `self._post` should be refactored as well. But didn't get about "code inserting", could you please clarify what did you mean?

Also, for now I don't know endpoint for this cases that returns not `json` data, so no changes regarding it is needed for now.

PR #98, opened 2020-06-18
fix/20259-fail_on_failed_machine

devProdigy:
## Description
Scheduler fails on DEV cause of inconsistency with Models building.
Aim of fix - is to skip invalid projects with models but to schedule valid one without total failing. 

**Please be aware that in one or two nearest sprints this logic will be deleted from Latigo and replaced by fetching proper data from Metadata API.**

* version 0.2.22 > 0.2.23

[Azure DevOps Link](https://dev.azure.com/EquinorASA/OMNIA%20Prevent/_workitems/edit/20259)

## Type of change

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

# Checklist:


app/latigo/gordo/model_info_provider.py
```diff
@@ -36,47 +38,64 @@ def get_model_data(
...
-
-            else:
-                logger.error(f"No client found for project '{project_name}', skipping")
+            project_machines = self._fetch_with_known_errors(
+                func=client._get_machines,
+                project_name=project_name,
+                machine_names=model_names,
+            )
+            if project_machines:
+                machines += project_machines
```
rkicz, line 47: i think machines.extend(project_machines) should be enough

devProdigy: Thanks, but we might have None there, so extend will raise an error in such case.

PR #99, opened 2020-06-25
Fix/20583 fail on failed machine

devProdigy:
## Description

Detailed description in the [Azure DevOps Link](https://dev.azure.com/EquinorASA/OMNIA%20Prevent/_workitems/edit/20583)

* disable of known error stack logging for failed Gordo metadata (not needed traceback for that error)
* use UTC timezone in scheduler Clock
* version 0.2.23 > 0.2.24

## Type of change

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

# Checklist:


app/latigo/scheduler/__init__.py
```diff
@@ -162,41 +164,23 @@ def perform_prediction_step(self):
...
-            logger.exception("Error occurred in scheduler")
-
-        return False
-
     def run(self):
         """Start the main loop."""
-        try:
-            self._run()
-        finally:
-            self.task_queue.close()
```
o-fedorov, line 186: Would you please add an explicit `self.task_queue.close()` for the run loop exit?

app/latigo/scheduler/__init__.py
```diff
@@ -162,41 +164,23 @@ def perform_prediction_step(self):
...
-
-    def _run(self):
         logger.info("Scheduler started processing")
-        done = False
-        start = datetime.datetime.now()
         if self.run_at_once:
-            done = self.on_time()
+            self._run()
```
o-fedorov, line 171: When `KeyboardInterrupt` is received here, it would be good to perform a graceful exit (at least call `self.task_queue.close()` explicitly).

app/latigo/scheduler/__init__.py
devProdigy: I checked and this part is throwing the exception cause `close` in called on delete of the `Consumer` which is happening before this place. So it called anyway.
Do you have another experience?

app/latigo/scheduler/__init__.py
devProdigy: I checked and this part is throwing the exception cause `close` in called on delete of the `Consumer` which is happening before this place. So it called anyway.
Do you have another experience?

app/latigo/scheduler/__init__.py
o-fedorov: There is no guarantee that `__del__` is executed prior to application exit. Also it’s better to close the connection as early as possible to ensure cleanup is done, even if a pod is forcibly terminated by Kubernetes some time after KeyboardInterrupt is received.

devProdigy:
```
Traceback (most recent call last):
  File "/Users/ocher2/Code/OMNIA/latigo/app/latigo/task_queue/kafka.py", line 81, in close
    self.producer.close()
AttributeError: 'cimpl.Producer' object has no attribute 'close'
Exception ignored in: <function KafkaTaskQueueSender.__del__ at 0x14537a0e0>
Traceback (most recent call last):
  File "/Users/ocher2/Code/OMNIA/latigo/app/latigo/task_queue/kafka.py", line 75, in __del__
  File "/Users/ocher2/Code/OMNIA/latigo/app/latigo/task_queue/kafka.py", line 81, in close
AttributeError: 'cimpl.Producer' object has no attribute 'close'
```

I've checked the [docs](https://docs.confluent.io/current/clients/confluent-kafka-python/#pythonclient-producer) and Producer does not have method `close` only Consumer has, so I suppose this [line](https://github.com/equinor/latigo/blob/master/app/latigo/task_queue/kafka.py#L81) should be deleted as well.
Or I missing smth?

o-fedorov: Sure, this line looks like a bug. Still `flush` method should be called prior to shutdown to ensure all produced messages are pushed to the queue.

devProdigy: Good point, added `close` call for all code block not to catch it twise.

app/latigo/scheduler/__init__.py
```diff
@@ -162,41 +164,26 @@ def perform_prediction_step(self):
...
+                self._run()
+
+            while self._is_ready:
+                try:
+                    self.continuous_prediction_timer.wait_for_trigger()
+                    self._run()
+                except KeyboardInterrupt:
+                    self._is_ready = False
+                except Exception:
+                    logger.exception("Error occurred in scheduler")
```
o-fedorov, line 181: Nit-pick: it would more readable to extract the code of this block to a separate method

PR #103, opened 2020-07-09
transfer from v1.5 to v1.6 of TS API.

devProdigy:
## Description

transfer from v1.5 to v1.6 of TS API.

* version 0.2.27 > 0.2.28;
* increase coverage to 68%;

**# TODO adjust configs with new version**

[Azure DevOps Link](https://dev.azure.com/EquinorASA/OMNIA%20Prevent/_workitems/edit/21254)

## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [X] Other

# Checklist:


app/latigo/time_series_api/sensor_data_provider.py
```diff
@@ -89,9 +89,9 @@ def get_data_for_range(
                 raise ValueError("'meta' was not found for name '%s' and asset_id '%s'", name, asset_id)
 
             item = _find_tag_in_data(meta, name)
-            tag_ids.append(item["id"])
+            tag_ids_names[item["id"]] = name
 
-        tags_data = self._fetch_data_for_multiple_ids(tag_ids=tag_ids, time_range=time_range)
+        tags_data = self._fetch_data_for_multiple_ids(tag_ids=list(tag_ids_names.keys()), time_range=time_range)
```
o-fedorov, line 94: More simple forms like `tag_ids=list(tag_ids_names)` and `tag_ids=tag_ids_names.keys()` should work.

app/latigo/time_series_api/sensor_data_provider.py
```diff
@@ -101,5 +101,8 @@ def get_data_for_range(
         if not tags_data:
             raise ValueError("No datapoints for tags where found.")
 
+        for tag_data in tags_data:  # add "tag_name" to data cause TS API does not return it anymore
+            tag_data["name"] = next(tag_name for tag_id, tag_name in tag_ids_names.items() if tag_id == tag_data["id"])
```
o-fedorov, line 105: Why not just this?
```suggestion
            tag_id = tag_data["id"]
            tag_data["name"] = tag_ids_names[tag_id]
```

app/latigo/types.py
```diff
@@ -56,6 +56,16 @@ class SensorDataSet:
...
     meta_data: typing.Dict = field(default_factory=dict)
 
+    def __eq__(self, other):
+        if self.time_range != other.time_range:
+            return False
+        if self.meta_data != other.meta_data:
+            return False
+        for i, series in enumerate(self.data):
+            if not series.equals(other.data[i]):
+                return False
```
o-fedorov, line 66: There should be a check for a case when `self.data` and `other.data` has a different number of elements.

Also current approach could be rewritten with `zip`. Putting everything together:

```suggestion
        if len(self.data) != len(other.data):
            return False

        if not all(s1.equals(s2) for s1, s2 in zip(self.data, other.data)):
            return False
```

Or:

```suggestion
        if len(self.data) != len(other.data):
            return False

        for self_series, other_series in zip(self.data, other.data)):
            if not series.equals(other_series):
                return False
```

setup.cfg
```diff
@@ -71,7 +71,7 @@ per-file-ignores=
     tasks.py:T001,WPS202
 
 [coverage:report]
-fail_under = 66
+fail_under = 68
```
o-fedorov, line 74: Great :)

app/latigo/time_series_api/sensor_data_provider.py
devProdigy: hm, that's simpler, seems like my brain over-engineered here after digging to deeply to Latigo..
thanks :)

app/latigo/types.py
devProdigy: Nice, this way is cleaner :+1:

app/latigo/time_series_api/sensor_data_provider.py
devProdigy: adjusted

PR #105, opened 2020-07-13
extend error message from ts API.

devProdigy:
## Description
Extend error message from TS API.
For now instead of full error from the API with explanations we see just generic error that describes status code.

* version 0.2.29 > 0.2.30

## Type of change

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [X] Other

# Checklist:


app/latigo/time_series_api/misc.py
```diff
@@ -65,8 +65,13 @@ def parse_request_json(res) -> typing.Dict:
...
     Raise:
         - HTTPError if response code is not 200.
+            Extend error with response text for more clarity.
     """
-    res.raise_for_status()
+    try:
+        res.raise_for_status()
+    except requests.HTTPError as e:
+        e.args = list(e.args) + [e.response.text]
+        raise e
```
o-fedorov, line 74: This looks like monkey patching. Altering an entity provided by third-party code is not safe. Why not just log out the response text?

devProdigy: Then it will be two separate log records and, I think, it's more convenient to have everything in one place. It's better not to make yourself doing additional steps in error finding, then it'll be more complicated.
Also, additional adjustments to dashboard, alerts and scrolling while debugging might be needed.

o-fedorov: Monkey patching should not be used in production code. No exclusions.

devProdigy: Can't see how is it monkey patching? This is what had to be done from the beginning, but we didn't take into account that `raise_for_status` do not carry error messages from the API.
There's no problem in such approach for me, if you want - please offer solution where it won't be separate line in log and will have consistency/simplicity in it.

devProdigy:
@o-fedorov 
From your logic - suppressing descriptive error from API is good practice and extending error with this really needed message is bad practice, how is that?
I disagree with opinion that current code change is bad practice. Also, usability and understandability is on the first place for any system and is above any practices, if code/log/app is hard to read/maintain/use then some of the principals were used in the wrong way. 
If you want and can - please offer solution where it won't be separate line in log and we will still have consistency/simplicity in it.

PR #106, opened 2020-07-20
21724 bring TS "facility" to code

devProdigy:
## Description

TS API has new filed that changed Latigo work a lot (in many places).
For now it's a bug and it's causing lots of errors.

Also, we have other tasks that will bring adjustments to current code and maybe TS API team will "bring" changes as well (so current approach probably is not final version)

* version 0.2.30 > 0.3.0;
* added tests and increased coverage.

[Azure DevOps Link](https://dev.azure.com/EquinorASA/OMNIA%20Prevent/_workitems/edit/21724)

## Type of change

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

# Checklist:


app/latigo/time_series_api/cache.py
```diff
@@ -20,36 +20,34 @@ class TagMetadataCache:
...
         if not res:
             raise Exception(f"Failed to store key '{key}' with value '{meta}' to cache.")
 
     @staticmethod
-    def _make_metadata_key(name: str, asset_id: str):
-        if not asset_id:  # better to check asset_id to make sure that consistency is followed
-            raise ValueError("'asset_id' can not be empty.")
-        return f"{asset_id}::{name}"
+    def _make_metadata_key(name: str, facility: typing.Optional[str]):
+        return f"{facility}::{name}"
```
o-fedorov, line 53: So, when `facility` is None, the key is `None::<name>`. Is it intended?

app/latigo/time_series_api/cache.py
```diff
@@ -20,36 +20,34 @@ class TagMetadataCache:
...
-            - asset_id: asset/project identifier. Example: "1000".
+            - facility: TS internal project identifier. Example: "1000".
         """
-        key = self._make_metadata_key(name, asset_id)
+        key = self._make_metadata_key(name, facility)
         dumped_meta = ujson.dumps(meta)
 
         res = self._cache.set(name=key, value=dumped_meta, ex=self.CACHE_TIME_TO_LIVE)
         if not res:
             raise Exception(f"Failed to store key '{key}' with value '{meta}' to cache.")
```
o-fedorov, line 49: I know it's out of the scope, but: why raising an exception? If the cache is not stored then there is simply no cache hit, Latigo still can finish its job.  It would be better to have to log an error and have some notification for it.

app/latigo/time_series_api/cache.py
devProdigy: Yes, as for me it's better to have None then any other hardcoded value.
For now None will be only in one scenario - when we want to fetch `common facility` for all tags by trying to get it from the first input tag of the model. In such case we don't know the `facility`. I'll discuss with TS API team (when they'll come back from vacation) options how we can get this data in other way, but for now it'll be like this. (talked about it with Alexis)

o-fedorov: I do not clearly realize how the cache is supposed to be used. Just, since `facility=None` is allowed for the cache, I want to ensure that there are no situation when the same tag name is used for different facilities, and due to that different values are overridden concurrently with the same key `None::<name>`. This may lead to issues that are hard to detect.

app/latigo/time_series_api/cache.py
o-fedorov: > if cache break then load on the service will be huge and it'll eventually start working with delays and loading TS API with calls that should be cached otherwise

It depends on what is better: to have delayed predictions, or to not have any predictions at all.

If the issue with cache is intermittent, it would slow down Latigo a little bit, and then when the cache starts working Latigo gets a chance to catch up. Usually temporary performance degradation is more preferable than complete deny of service. But again, it's out of the scope of current PR.

app/latigo/time_series_api/cache.py
devProdigy: I understood your point, this is open question, so I created a panel in dashboard to check for the "duplicates" and have to sync with TS API team about it cause it appeared that they have them and will clean the data. We don't have any other place from where and way to take `facility`. So if any duplicates will be returned error will be raised and we'll see if such approach is applicable here.

app/latigo/time_series_api/cache.py
devProdigy: I'll merge this PR, I think we might come back to this question later.

PR #110, opened 2020-07-24
21443 minimize calls to Redis

devProdigy:
## Description
21443 minimize calls to Redis

* version 0.3.1 > 0.3.2;
* removed additional calls to fetch Input Tag IDs;
* removed some unused code.

[Azure DevOps Link](https://dev.azure.com/EquinorASA/OMNIA%20Prevent/_workitems/edit/21443)

## Type of change

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

# Checklist:

- [x] I have added tests that prove my fix is effective or that my feature works (if needed)


app/latigo/metadata_api/metadata_storage_provider.py
```diff
@@ -49,13 +45,13 @@ def put_prediction_metadata(
             # example: operation: "start", "end", "model-input"; tag_name:  "1903.R-29TT3018.MA_Y"
 
             if operation == MODEL_INPUT_OPERATION:
-                tag_time_series_id = input_time_series_ids[tag_name]
+                tag_time_series_id = prediction_tags_data[col].ts_tag_id
```
o-fedorov, line 48: I wonder if it makes sense extracting a variable for `prediction_tags_data[col]`

app/latigo/metadata_storage/__init__.py
```diff
@@ -1,65 +1,27 @@
...
 
     if "metadata_api" == provider_type:
         from latigo.metadata_api.metadata_storage_provider import MetadataAPIMetadataStorageProvider
 
         provider = MetadataAPIMetadataStorageProvider(config)
-    elif "mock" == provider_type:
-        provider = MockMetadataStorageProvider(config)
     else:
-        provider = DevNullMetadataStorageProvider(config)
+        raise ValueError(f"'{provider_type}' is not supported.")
```
o-fedorov, line 26: It looks to me that you may get rid of all these "provider_factory" stuff and just build the instances explicitly. It would improve the readability of the code.

devProdigy: Agree, I'm thinking to add task to one of the next sprints to get rid of such "interfaces and factories". I think we don't need them in Latigo. Instead simple instances should be populated (like we did for Redis cache).
Also, config file should be "cleaned" as well.
It's a lot of work, so I hope we'll can get time on it after/along with other priority tasks.

app/latigo/metadata_api/metadata_storage_provider.py
devProdigy: We can remove it and access to it directly. Adjusted

app/latigo/time_series_api/client.py
```diff
@@ -190,6 +190,42 @@ def _create_id_if_not_exists(
...
+        tags_metadata = self._tag_metadata_cache.get_multiple_metadata(
+            tag_names=list(tag_names_descriptions), facility=facility
+        )
+
+        for tag_name, metadata in tags_metadata.items():
+            if not metadata or not metadata["data"]["items"]:
+                meta = self.replace_cached_metadata_with_new(
+                    tag_name=tag_name, facility=facility, description=tag_names_descriptions[tag_name]
+                )
+                tags_metadata[tag_name] = meta
```
o-fedorov, line 224: I bet for these ones `mset` could be used as well.

devProdigy: Yes, this might be the next step of improvement.

PR #116, opened 2020-08-25
21750 create/fetch + cache TS tags as bulk.

devProdigy:
## Description
Use bulk methods to:
- create/fetch TS tags from TS API;
- set tag to cache. 

Generalize caching storing and returning data from TS API without unneeded data.
Version 0.3.6 -> 0.4.0.

**TBD**: clear cache in Redis after deploy to the ENVs.

[Azure DevOps Link](https://dev.azure.com/EquinorASA/OMNIA%20Prevent/_workitems/edit/21750)

## Type of change

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

# Checklist:


app/latigo/time_series_api/cache.py
```diff
@@ -63,6 +63,38 @@ def set_metadata(self, name: str, facility: Optional[str], meta: Dict):
...
+            result of the set operations as bools.
+        """
+        pipe = self._cache.pipeline()
+        for meta in tags_metadata:
+            dumped_meta = ujson.dumps(meta)
+            key = self._make_metadata_key(meta["name"], facility)
+            pipe.set(name=key, value=dumped_meta, ex=self.CACHE_TIME_TO_LIVE)
+        return pipe.execute()
+
+    def delete_all_keys(self):
```
o-fedorov, line 94: Doing this under high load will immediately kill the performance if not the service itself. Why it is needed?

app/latigo/time_series_api/client.py
```diff
@@ -356,15 +403,14 @@ def get_facility_by_tag_name(self, tag_name: str) -> str:
...
-        ts_tag = self.get_only_item_from_metadata(metadata=res, tag_name=tag_name)
+        ts_tag = self.get_meta_by_name(name=tag_name)
         facility = ts_tag["facility"]
         if not facility:
             raise ValueError(f"tag 'facility' is empty for tag name '{tag_name}' with data: {ts_tag}")
         return facility
 
     @staticmethod
-    def get_only_item_from_metadata(metadata: dict, tag_name: str) -> dict:
+    def get_only_item_from_metadata(metadata: dict, tag_name: typing.Optional[str] = None) -> dict:
```
o-fedorov, line 413: The tag is passed but not used for any reason other than informational.  You should remove it completely.

devProdigy: Without it logs will be useless and debugging is very hard. There's one case when it won't be passed but in all other cases it's very important to have it in the logs

app/latigo/time_series_api/client.py
```diff
@@ -312,26 +365,20 @@ def store_multiple_datapoints(self, datapoints_to_store: typing.List[typing.Dict
...
-            raise ValueError(f"Could not replace cached metadata for {tag_name}/{facility}/{description}")
+        tags_metadata = self._get_or_create_tags([
+            self._make_tag_metadata(name=tag_name, description=description, facility=facility)
+        ])
 
         # set in cache with new value
-        self._tag_metadata_cache.set_metadata(tag_name, facility, meta)
-        return meta
+        self._tag_metadata_cache.set_metadata(tag_name, facility, tags_metadata[0])
+        return tags_metadata[0]
```
o-fedorov, line 381: A variable with a meaningful name could be used instead of `tags_metadata[0]`

app/latigo/time_series_api/cache.py
devProdigy: this won't be used in the production, it's more like code that we can use locally - not to write this functionality all the time.
I don't have any functionality in my mind that will require doing it in live :)

app/latigo/time_series_api/client.py
devProdigy: sorry, didn't get it. Do you mean to create additional variable to fetch first `tag_metadata` from `tags_metadata`? 
There's no chance that more than one element will be returned, so it's just basically what it is - first element.

app/latigo/time_series_api/client.py
o-fedorov: Then just add logs, and inject the tag name to them with `pylogctx` somewhere in the upstream code (if it is not done yet).

app/latigo/time_series_api/cache.py
o-fedorov: Then do not keep it here, or somebody may accidentally add a call to this method in the future.  Isuues like this are realy hard to debug. Just move clearing the cache to testing code.

app/latigo/time_series_api/client.py
o-fedorov:
```suggestion
        current_tag_metadata = tags_metadata[0]
        self._tag_metadata_cache.set_metadata(tag_name, facility, current_tag_metadata)
        return current_tag_metadata
```

app/latigo/time_series_api/client.py
devProdigy: Then people should remember to add tag_name to `pylogctx` each time they call this function. And the call-chain might be to long for person to remember about that. So it'll be missed. Also multiple calls could be done and then updating and clearing the `pylogctx` might be overhead for this case.

P.S. it was my mistake, it shouldn't be `Optional` with None, it'd be mandatory. Changed it.

app/latigo/time_series_api/cache.py
devProdigy: what do you mean under the "testing code"?
Also, this is the fastest way to delete keys from the database in Redis, so it's not the issue from what I know. The faster way for the calling-code might be to add `async` to the redis-function call.

devProdigy: anyway, since it's not used now - it might be deleted from the codebase.

o-fedorov (approved):

rkicz (approved):

PR #121, opened 2020-09-25
26419 do not create input tag ever

devProdigy:
## Description
We should not create input tag ever.
Divide logic for input and output tags in the system:
* input tag - only fetch and throw exception if not found (in bulk);
* output tag - fetch or create if not found.

Additionally:
* add tests;
* added forgotten auth check for TS API on the Latigo start;
* increase amount of cache life for 2 days (data is showing more stability, so it could be increased);
* version 0.5.0.

[Azure DevOps Link](https://dev.azure.com/EquinorASA/OMNIA%20Prevent/_workitems/edit/26419)

## Type of change

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] CI related changes
- [X] Tests
- [ ] Other

# Checklist:


app/latigo/time_series_api/client.py
```diff
@@ -171,13 +176,25 @@ def _get_metadata_from_api(self, name: str, facility: typing.Optional[str] = Non
...
+
+        for batch in get_batches(tag_names, batch_size=max_tag_names_in_one_request):
+            names_to_fetch = "&".join((f"name={name}" for name in batch))
+            url = f"{self.base_url}?{names_to_fetch}{f'&facility={facility}' if facility else ''}"
+            if len(url) >= max_url_chars_amount:
+                raise ValueError(
+                    f"Length of the url to get Tags exceeded maximum allowed length and has '{len(url)}' chars."
+                )
+
+            res = self._get(url)
```
o-fedorov, line 193: There is no need to compound a URL stirng, `requests` accepts a list with `params` argument:
```suggestion
            params = {"name": batch}
            if facility:
                params["facility"] = facility
            res = self._get(url, params=params)
```

app/latigo/time_series_api/client.py
```diff
@@ -227,24 +247,49 @@ def get_or_create_multiple_tags(self, tag_names_descriptions: typing.Dict[str, s
...
-                tags_to_get_or_create.append(self._make_tag_metadata(
-                    name=tag_name, description=tag_names_descriptions[tag_name], facility=facility,
-                ))
+                tag_not_in_cache = next(t for t in tags_info if tag_name == t.ts_tag_name)
+
+                # 'input' and 'output' tags should treated differently:
+                # Input tags should not be created ever, that's why they should have "raise_error_if_missing" == True,
+                # so for input tags only "fetch" method should be used.
+                # Output tags use getOrAdd method that creates tag if it doesn't exist.
+                if tag_not_in_cache.raise_error_if_missing:
```
o-fedorov, line 263: Having to explain the attribute in a code that *uses* it is a clear sigh of bad design. Also I see that one have to explain this attribute *everywhere* it is used.

A better solution would be to distinguish the fact that the tag is missing, and the fact that it is an input or an output tag.

app/latigo/time_series_api/client.py
```diff
@@ -227,24 +247,49 @@ def get_or_create_multiple_tags(self, tag_names_descriptions: typing.Dict[str, s
...
+            not_cached_tags.extend(self._get_or_create_tags(output_tags_to_get_or_create))
+        if input_tags_names_to_fetch:
+            fetched_input_tags = self._get_multiple_tags_metadata(
+                tag_names=input_tags_names_to_fetch, facility=facility
+            )
+            missing_input_tags = set(input_tags_names_to_fetch) - set(
+                input_tag_metadata["name"] for input_tag_metadata in fetched_input_tags
+            )
+            if missing_input_tags:
+                raise ValueError(f"Following input tags are missing in the TS API: {';'.join(missing_input_tags)}")
```
o-fedorov, line 283: Should it be raised at the very top of the method? Ideally when before doing anything to output tags.

devProdigy: this method is the only place where it can be found for the sake of the logic of Latigo. 
It fetches the tags and not to check that input tag is missing on the all levels - it's safer and avoid code duplicates.
Basically it's the first main thing that app will do.

app/latigo/time_series_api/client.py
devProdigy: You're right, `requests` does accept params, I thought that it'd be useful to have check that will throw an error with some details if somehow tags be named very largely, url will be too long and it'll make system crush. 
Now I think that we'll get the error from the `requests` anyway, so length of the url might be investigated with reproducing same model prediction from the logs of the error.
I'll adjust this.

app/latigo/time_series_api/client.py
devProdigy: explanation was added to 100% be sure why it was done in this particular case, there might be no need in so detailed explanation whatsoever :) 
Is your suggestion to rename `raise_error_if_missing` to `is_input_tag`? If yes then I like to have more broad option to determine any of the tags as `raise_error_if_missing` in case of need. Maybe there be any other edge cases when some of the input tags might be created and some of the output tags are not.

app/latigo/time_series_api/client.py
devProdigy: ok, additionally what I wrote above I think I got your point: "First check the input tags and if no errors - then output tags".
Switched them in places.

app/latigo/time_series_api/client.py
o-fedorov: First of all, the problem is that data should not be mixed with business logic. A dataclass can't say "Raise an error if I am missing", without providing a way to do so, which means that either *the logic should be encapsulated to the dataclass*, or, more preferred, the decision to raise an error should be deferred to *the place where it is possible to check if the entity is missing*. That is, there is only one place where the tag is both checked for existence and where a decision for fetching missing ones is made.

Second, here in the code `raise_error_if_missing` is used to determine input tags.  It's a logical error, since, as you mentioned above, `raise_error_if_missing` has more broad meaning than just "input tags". And if some developer decide to set `raise_error_if_missing` to True for output tags, an assumption that `input_tags_names_to_fetch` contains input tags will be wrong. Also the logic for it would be spread across the codebase, which is error-prone.

Third, `PredictionTSTagInfo` has a global usage, while `raise_error_if_missing` attribute makes sense only for this particular method of code. That is:

* There is no value for other parts of the code to know that an error should be raised if the tag is missing.
* `raise_error_if_missing` is simply ignored if `get_or_create_multiple_tags` not called.

**To sum up**:

* Apply [Information hiding](https://en.wikipedia.org/wiki/Information_hiding) principle to provide only the information that makes sense to the `PredictionTSTagInfo` dataclass, not trying to predict the ways business logic will handle it.
* [Encapsulate](https://en.wikipedia.org/wiki/Encapsulation_(computer_programming)) the tags checking and validation in a single place, do not leak its implementation details to `PredictionTSTagInfo`, or to the methods outside current one.

devProdigy: Thanks for explanations.
Since we have only two types `input` and `output` tags, I'll make it simple bool value, so the logic will be encapsulated and decision on any particular logic in other pieces of code will be made independently.

app/latigo/time_series_api/misc.py
```diff
@@ -104,7 +104,16 @@ def find_in_time_series_resp(res, x):
...
+        description: Description will be saved to new Tag if no Tag be found.
+        ts_tag_id: internal TS API tag ID.
+        is_input_tag: if True this tag has is 'input-tag'.
+    """
+
     ts_tag_name: str
-    description: str
+    description: str = ""
     ts_tag_id: str = ""
+    is_input_tag: bool = False
```
o-fedorov, line 119: I see that `is_input_tag` is determined by `ts_tag_name`.  Moving related logic here would protect from users forgetting to set `is_input_tag` value:

```
from dataclasses import dataclass, field
...


@dataclass
class PredictionTSTagInfo:
    """Info about Tag where prediction datapoints will be written.
    Attrs:
        ts_tag_name: Name will have new Tag if no Tag be found.
        description: Description will be saved to new Tag if no Tag be found.
        ts_tag_id: internal TS API tag ID.
        is_input_tag: if True this tag is 'input-tag'.
    """

    ts_tag_name: str
    description: str = ""
    ts_tag_id: str = ""
    is_input_tag: bool = field(init=False)

    def __post_init__(self):
        self.is_input_tag = False if OUTPUT_TAG_SEPARATOR in self.ts_tag_name else True
```

More details about [post-init](https://docs.python.org/3/library/dataclasses.html#post-init-processing)

o-fedorov (dismissed):

app/latigo/time_series_api/misc.py
devProdigy: Good point, from the beginning I though that it might be another cases except that one with the "delimiter" but seems like it's the only case for now.
Implemented this suggestion.

o-fedorov (approved):

PR #126, opened 2020-10-13
27271 add retries on scheduler on error.

devProdigy:
## Description
Add Scheduler ability to retry attempt to schedule predictions for execution if any unexpected error happened.
There'll be 2 re-tries with sleep for 5 mins between each.

* update Gordo version 0.68.0 > 0.69.2;
* version 0.5.3 > 0.5.4.

[Azure DevOps Link](https://dev.azure.com/EquinorASA/OMNIA%20Prevent/_workitems/edit/27271)

## Type of change

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

# Checklist:


app/latigo/utils.py
```diff
@@ -346,3 +349,32 @@ def get_batches(iterable, batch_size: int = 10) -> typing.Generator:
...
+                    return func(*args, **kwargs)
+                except Exception:
+                    logger.exception(
+                        "Error occurred in scheduler. Repeats left: %s. Sleeping between calls %s seconds.",
+                        repeats,
+                        sleep_time,
+                    )
+                    repeats -= 1
+                    time.sleep(sleep_time)
+
```
devProdigy, line 377: How do you think: should we re-raise exception here?
So maybe after the last re-try instead of logging the exception just re-raise it so it'll be handled on the top level?

rkicz: i would go with re-raise and maybe exponential backoff retry [like here](https://github.com/equinor/omnia-prevent-mdapi/pull/198/files#diff-338b5d63d806237bd963c40ccb0c96560823e927e912f41c36f56191691d9646R11)

devProdigy: Thanks, I like idea of incremental sleep 👍 
But in this case sleep should be much longer than in example. Code was adjusted.

rkicz (approved):

PR #130, opened 2020-11-11
Feature/rivan 29482 update docker workflow steps

RollerKnobster:
## Description

This PR splits the existing `main.yml` Github workflow into CI and Docker related ones, also updating them with the usage of actions instead of direct bash commands.

[Azure DevOps Link](https://dev.azure.com/EquinorASA/OMNIA%20Prevent/_workitems/edit/29482)

## Type of change

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes
- [x] CI related changes
- [x] Documentation content changes
- [ ] Tests
- [ ] Other

# Checklist:

- [x] I have made corresponding changes to the documentation / README (if needed)


.github/workflows/docker.yml
```diff
@@ -0,0 +1,174 @@
+name: Docker
+
+on:
+  pull_request:
+    branches:
+      - master
```
devProdigy, line 6: will it build docker image for each PR?
if yes, then as I told on the call - cause of the long time of build I'm against building docker on PRs, only for master branch (after merge). It'll slow down development.
This is not the ideal solution and should be changed after moving Gordo cliet to separate library.

README.md
```diff
@@ -221,16 +222,51 @@ Do not forget update already installed packages:
...
+full version for a given commit during a pull request (`v1.0.1dev5`).
+
+The format of each part is configurable and may be changed to one's likings. For our purposes, they're set to `<major>.<minor>.<patch>`
+for the main part, `dev` for the delimiter part, and `<increment>` for the increment part. The "increment" stands for the number of commits for a specific version
+since the last commit that could set the highest version, starting from either the latest release tag or from the start of the log tree.
+
+A link to the GitHub release creation page is generated in the outputs of the Semantic Version step of Docker CI workflow,
+setting the tag to follow the format `<tag_prefix><version>`. `<tag_prefix>` is also a configurable variable, set to `v` for our purposes.
+
+Many examples and details on the topics can be found by following the links, but in general, every commit should follow a couple of rules:
```
devProdigy, line 241: sorry, I was confused what links?)
```suggestion
Many examples and details on the topics can be found by following the links above, but in general, every commit should follow a couple of rules:
```

README.md
```diff
@@ -221,16 +222,51 @@ Do not forget update already installed packages:
...
+
+BREAKING CHANGE: `extends` key in config file is now used for extending other config files
+```
+
+Given this is the first commit since the release ref, the version after merging shall be set as `2.0.0dev0`.
+This makes sense due to the fact that a breaking change is present in a commit.
+A link to make a release with tag `v2.0.0` is going to be created.
+
+After pushing to master or creating a tag, the new image with the appropriate version is going to be 
+created and pushed to the registry.
```
devProdigy, line 260: Sorry, didn't understand this part.
So what is the common logic of the CI-docker-push?
Maybe some short summary/example could be added to the beginning "### Versioning" section?
Like:
```
1. create PR with `BREAKING FEATURE:`\`feat`\`fix` in the any of commit's message (here some explanations how logic will determine what to increment if be 3 commits with all 3 options).
2. Image for DEV: ....
3. Image for PROD: After merge to Master...
etc.
```

.github/workflows/docker.yml
```diff
@@ -0,0 +1,174 @@
...
+        TAG=$([[ "$GITHUB_REF" == "refs/heads/master" ]] && echo "${{ steps.versioning.outputs.tag }},latest" || echo "${{ steps.versioning.outputs.tag }}")
+        EOF
+
+    - name: List environment info
+      run: |
+        make show-env
+        make info
+
+    - name: Set up QEMU
+      uses: docker/setup-qemu-action@v1
```
devProdigy, line 82: could you please explain why is this needed?

.github/workflows/docker.yml
```diff
@@ -0,0 +1,174 @@
...
+        REGISTRY: ${{ secrets.DOCKER_REGISTRY }}
+        REPO: ${{ secrets.DOCKER_REPO }}
+      with:
+        context: .
+        file: ./Dockerfile
+        tags: |
+          ${{ env.BASE_IMAGE }}
+          ${{ env.BASE_IMAGE }}:${{ env.VERSION }}
+        labels: |
+          ${{ steps.docker_meta.outputs.labels }}
```
o-fedorov, line 105: Is it really needed?

o-fedorov (dismissed):

.github/workflows/docker.yml
RollerKnobster: I'll comment out the `pull_request` part then and leave a TODO to bring it back when Gordo client is separated.

README.md
RollerKnobster: I meant the ones in the beginning of the section. Will clarify.

README.md
RollerKnobster: Will elaborate.

.github/workflows/docker.yml
RollerKnobster: To be honest, we probably don't need QEMU. `buildx` is needed in order to enable the `build-push-action` to function properly, as with moving to `v2` it moved to using `docker buildx` toolset for building images. But we certainly don't need support for multiple platforms. Will remove.

.github/workflows/docker.yml
RollerKnobster: I guess we could do it all manually. This is needed to automate at least some part of the standard set of [OCI labels](https://github.com/opencontainers/image-spec/blob/master/annotations.md), but we could do without that.

.github/workflows/docker.yml
devProdigy: did you comment it?)

.github/workflows/docker.yml
```diff
@@ -0,0 +1,170 @@
+name: Docker
+
+on:
+#  We should try and build the image on every pull request, but only after we clear out the Gordo dependencies
```
o-fedorov, line 4: Could you please create an Azure DevOps task for it (if there is no already) and place a link or ID of it here?

.github/workflows/docker.yml
```diff
@@ -0,0 +1,170 @@
...
+        docker pull -q alpine > /dev/null
+        docker run --rm -v "$(pwd)/trivydb/":/root/.cache/ $(docker images alpine --format '{{.Repository}}:{{.Tag}}' | head -1) chmod 777 -R /root/.cache/
+
+    - name: Push images to dev registry
+      if: ${{ !startsWith(github.ref , 'refs/pull/') }}
+      env:
+        USERNAME: ${{ secrets.DOCKER_USERNAME }}
+        PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
+        REGISTRY: ${{ secrets.DOCKER_REGISTRY }}
+        REPO: ${{ secrets.DOCKER_REPO }}
```
o-fedorov, line 140: For better readability it's better to have REGISTRY and REPO configured in this file, not in secrets.

BTW, now that we've had problems with docker registry credentials multiple times I start to thinking of exposing DOCKER_USERNAME as well, so that when credentials should be updated it's easier to find the client that has access to the registry. That is, I think only `DOCKER_PASSWORD` and `PROD_DOCKER_PASSWORD` should be in the secrets.

Still, it's not a requirement for current PR.

o-fedorov (dismissed):

.github/workflows/docker.yml
RollerKnobster: I am all for that actually.

.github/workflows/docker.yml
RollerKnobster: Absolutely.

devProdigy (approved):

PR #134, opened 2020-11-20
feat: 29829 allow different tag`s facilities in one model.

devProdigy:
## Description

Now in one models we might have tags with different facilities (got the common one as before).

* update requirements;

[Azure DevOps Link](https://dev.azure.com/EquinorASA/OMNIA%20Prevent/_workitems/edit/29829)

## Type of change

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Tests
- [ ] Other

# Checklist:


app/latigo/gordo/misc.py
```diff
@@ -67,3 +70,26 @@ def print_client_debug(client: Optional[Client]):
...
+    Returns:
+        tag`s facility.
+    """
+    letters_to_cut_from_asset = 4
+
+    if not tag_name and not sensor_tag:
+        raise ValueError("'tag_name' or 'sensor_tag' should be populated.")
+
+    sensor_tag = sensor_tag if sensor_tag else normalize_sensor_tags([tag_name])[0]
+    return sensor_tag.asset[:letters_to_cut_from_asset]
```
koropets, line 95: A better solution here would be taken the first numbers from `asset` and if there are no numbers there just keeping `asset` as it is

devProdigy: is there such possibility that facility won't has numbers? (like 1000)
I thought that asset will always start from 4 numbers - facility.

koropets: No, but it might be in the future

koropets: For our future clients

devProdigy: I offer not to rely on it now, cause we don't know how it'll be and such change should be added explicitly and such facilities should supported by TS API (or any other provider). So it probably be more complex than just trying to get any with number.

rkicz (approved):

RollerKnobster (approved):

PR #140, opened 2020-11-26
Feat: 30088 enable static code/analysers on CI

RomanRusyn:
* enabled flake8, black for app/bin on CI
* cleaned app/bin with flake8, black

## Description

Summary of the changes / which issue is fixed / etc.

[Azure DevOps Link](https://dev.azure.com/EquinorASA/OMNIA%20Prevent/_workitems/edit/17188)

## Type of change

- [ ] Bugfix
- [ ] Feature
- [x] Code style update (formatting, local variables)
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [x] CI related changes
- [ ] Tests
- [ ] Other

# Checklist:


code_quality/Makefile
```diff
@@ -7,10 +7,10 @@ TESTS_DIR="${ROOT_DIR}/../tests"
 
 h: help
 
-all: black flake mypy #pylint is broken, disabled for now
+all: black flake #pylint is broken, disabled for now. mypy disabled for testing
 
 flake:
-	flake8 --ignore=E731,W503,W504,E501,E265,C0301,W1202,W1203 --max-complexity 10 --exclude build,junk --exit-zero "${ROOT_DIR}/app"
+	flake8 --ignore=E731,W503,W504,E501,E265,C0301,W1202,W1203 --max-complexity 10 --exclude build,junk --exit-zero ../app/bin
```
devProdigy, line 13: does [ROOT_DIR](https://github.com/equinor/latigo/blob/master/Makefile#L2) variable from makefile not work? Or [APP_DIR](https://github.com/equinor/latigo/blob/master/Makefile#L3)
so it will look like `"${APP_DIR}/bin"`.

devProdigy: Please move all settings for the flake to proper config files. Like this [one](https://github.com/equinor/latigo/blob/master/setup.cfg). It already has configs for Flake.

code_quality/Makefile
```diff
@@ -22,8 +22,8 @@ mypy:
 	mypy --ignore-missing-imports "${TESTS_DIR}"  || true
 
 black:
-	black -l 120 -t py37 "${APP_DIR}"
-	black -l 120 -t py37 "${TESTS_DIR}"
+	black -l 120 -t py37 ../app/bin  # path for testing phase1
```
devProdigy, line 25: Please see comment about path variable for Flake command.

Also, please move all settings for the black to proper config file that Black supports.

code_quality/Makefile
```diff
@@ -7,10 +7,10 @@ TESTS_DIR="${ROOT_DIR}/../tests"
 
 h: help
 
-all: black flake mypy #pylint is broken, disabled for now
+all: black flake #pylint is broken, disabled for now. mypy disabled for testing
```
devProdigy, line 10: Please remove `pylint` from the makefile and other places in code. We won't use it.

.github/workflows/ci.yml
```diff
@@ -57,3 +57,6 @@ jobs:
         METADATA_API_AUTHORITY_HOST_URL: ${{ secrets.METADATA_API_AUTHORITY_HOST_URL }}
         METADATA_API_BASE_URL: ${{ secrets.METADATA_API_BASE_URL }}
         METADATA_API_APIM_KEY: ${{ secrets.METADATA_API_APIM_KEY }}
+    - name: Run all checks
```
devProdigy, line 60: tests run much longer then code checks, so it should be step before `name: Run All Tests`

devProdigy (changes requested):

code_quality/Makefile
```diff
@@ -7,23 +7,19 @@ TESTS_DIR="${ROOT_DIR}/../tests"
...
-all: black flake mypy #pylint is broken, disabled for now
+all: black flake  # mypy disabled for testing
 
 flake:
-	flake8 --ignore=E731,W503,W504,E501,E265,C0301,W1202,W1203 --max-complexity 10 --exclude build,junk --exit-zero "${ROOT_DIR}/app"
-
-pylint:
-	pylint --exit-zero --rcfile pylintrc --output-format=colorized "${APP_DIR}"
-	pylint --exit-zero --rcfile pylintrc --output-format=colorized "${TESTS_DIR}"
+	flake8 --exit-zero "${APP_DIR}/bin"
```
devProdigy, line 13: why do we need ` --exit-zero`? It means "Exit with status code "0" even if there are errors." but the main idea is to enable flake and fix errors for 3 files in `bin` dir.

setup.cfg
```diff
@@ -36,28 +36,67 @@ ignore =
...
+    # Missing "Yields" in Docstring
+    DAR301
+    # Missing exception(s) in Raises section
+    DAR401
+    # Found `%` string formatting
+    WPS323
+    # Found an unnecessary blank line before a bracket - handled by Black
+    WPS355
+    # Missing docstring in public package
+    D104
```
devProdigy, line 80: I don't like the idea to ignore such errors for all files.
Most of them are reasonable warnings, so if it's really the case for some particular place is code - then we should ignore them per files and not to all project.

Is there any common errors that should be handled for all files cause of specific of black + flake?

setup.cfg
```diff
@@ -36,28 +36,67 @@ ignore =
...
 
 
 per-file-ignores=
```
devProdigy, line 83: please add task to the same Story to clean "global ignores" in the config files.

app/bin/executor.py
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env python
+"""Executor runner module."""
```
devProdigy, line 2:
```suggestion
"""Entry point to run Executor."""
```

app/bin/scheduler.py
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env python
+"""Scheduler runner module."""
```
devProdigy, line 2:
```suggestion
"""Entry point to run Scheduler."""
```

.github/workflows/ci.yml
```diff
@@ -46,6 +46,10 @@ jobs:
         pip install -r app/requirements.txt
         pip install -r app/test_requirements.txt
 
+    - name: Run all checks
```
devProdigy, line 49:
```suggestion
    - name: Run code quality checks
```

devProdigy (changes requested):

setup.cfg
RollerKnobster: I could agree on the idea of not having the docstrings warnings ignored (DARxxx), but otherwise all of the other errors are either contradicting `black` formatting or limit the usage of genuine self-documented and useful pieces.

devProdigy: agree with what you offered.

Makefile
```diff
@@ -18,9 +18,24 @@ ifeq ($(strip $(TRIVY_TIMEOUT)),)  # add default timeout for Trivy if not passed
 TRIVY_TIMEOUT:="10m"
 endif
 
-.PHONY: all code-quality tests_all tests_unit tests_integration set-env setup up rebuild-req
+.PHONY: all code-quality tests_all tests_unit tests_integration setup up all flake mypy coverage black
```
RollerKnobster, line 21:
```suggestion
.PHONY: all flake mypy coverage black tests_all tests_unit tests_integration set-env setup up rebuild-req
```

Makefile
```diff
@@ -18,9 +18,24 @@ ifeq ($(strip $(TRIVY_TIMEOUT)),)  # add default timeout for Trivy if not passed
...
 
-all: help
+flake: ## Run flake
+	flake8 "${APP_DIR}/bin"
+
+mypy: ## Run mypy
+	# mypy is sometimes too sensitive so we allow failure
+	mypy --ignore-missing-imports "${APP_DIR}" || true
+	mypy --ignore-missing-imports "${TESTS_DIR}"  || true
```
RollerKnobster, line 29: Please remove `--ignore-missing-imports` from these calls and instead change `ignore-missing-imports` to `true` in `mypy.ini`

RollerKnobster: Also change this step to run `mypy` for the `bin` folder and enable it, removing `|| true`

Makefile
```diff
@@ -18,9 +18,23 @@ ifeq ($(strip $(TRIVY_TIMEOUT)),)  # add default timeout for Trivy if not passed
...
+	mypy "${APP_DIR}/bin"
+
+black: ## Run black
+	black "${APP_DIR}/bin"
+
+
+coverage: ## Run coverage
+	python -m pytest --junitxml=junit-results.xml --cov="${APP_DIR}" --cov-report=xml --cov-report=html unit
+
+#	autopep8 -r -i -j 0 --max-line-length 999 -a -a -a ./
```
devProdigy, line 37: could you please explain what is this line for?

Makefile
```diff
@@ -43,7 +57,7 @@ info:  ## Show info about env and used variables
 	@echo "DOCKER_PASSWORD=NOT SHOWN"
 
 code-quality:  ## Run code quality tools
-	cd "${CODE_QUALITY_DIR}" && make all
+	make black flake  # mypy disabled for testing
```
devProdigy, line 60: I think this will produce additional line of the output like `make black flake`.

```suggestion
code-quality: black flake  ## Run code quality tools
```
OR
```suggestion
code-quality:  ## Run code quality tools
	@make black flake  
```

devProdigy (approved):
Nice work 👍

